PR #4, opened 2019-07-15
ITE-2: Using TUF and in-toto to build compromise-resilient CI/CD

trishankatdatadog:
We discuss a standard for using The Update Framework (TUF) as a higher-level protocol for securely distributing in-toto layout and link metadata as well as the software packages described in the in-toto metadata. Assuming that critical signing keys are kept offline from the CI/CD pipeline, combining TUF with in-toto provides the pipeline with a desirable property called _compromise-resilience_: that is, even if the pipeline is compromised anywhere between developers and end-users, then attackers should not be able to cause end-users to install malicious versions of packages that were never released by developers.


trishankatdatadog:
@jhdalek55 Addressed your early comments, thanks!

@SantiagoTorres @JustinCappos Please send feedback, thanks!

ITE/2/README.adoc
```diff
@@ -0,0 +1,977 @@
...
+offline keys, where m >= 1, and n is the number of currently trusted
+developers. It is OPTIONAL but RECOMMENDED to protect these keys using
+hardware such as https://github.com/DataDog/yubikey[YubiKeys].
+. The "`packager`" step is performed by the CI/CD pipeline which
+transforms this source code into packages (e.g., tarballs, binaries).
+Typically, it receives source code from the "`developer`" step, and
+outputs packages as products. It SHOULD use (1, 1) online keys.
+
+The in-toto root layout itself MUST be signed by administrators using
+(m, n) offline keys, where n >= 2, and m >= ceiling(n / 2). This means
```
JustinCappos: Why does this need to be at least n/2?

trishankatdatadog: I'm thinking majority vote for the in-toto root layout, too, no, just like the TUF root role? Any reason why this should _not_ be the case?

trishankatdatadog:
One idea is to make ITE-2 all about the basic security model

And ITE-3 about the Datadog Agent integrations

What do you all think?

jhdalek55:
Trishank's suggestion of keeping ITE-2 more general and covering the specific Datadog integration more general makes sense to me.

ITE/2/README.adoc
JustinCappos: The majority is just an artificial way to do this.  If an org feels likely their keys may be lost but are less likely to be compromised / coerced, they may want a lower threshold.  I don't think there is anything magical about >50%, so I don't think we should make it a MUST.

ITE/2/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+. The "`developer`" step is performed by a developer who signs source
+code. Typically, it receives no materials, and outputs source code as
+products. Its link metadata MUST be signed by developers using (m, n)
+offline keys, where m >= 1, and n is the number of currently trusted
+developers. It is OPTIONAL but RECOMMENDED to protect these keys using
+hardware such as https://github.com/DataDog/yubikey[YubiKeys].
+. The "`packager`" step is performed by the CI/CD pipeline which
+transforms this source code into packages (e.g., tarballs, binaries).
+Typically, it receives source code from the "`developer`" step, and
+outputs packages as products. It SHOULD use (1, 1) online keys.
```
JustinCappos: Talking about the rationale here (why SHOULD, when to use other values, etc.) may be helpful.

ITE/2/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+    }
+    ...,
+  }
+}
+----
+
+The targets role signs targets metadata about the in-toto root layout,
+and all of the public keys used to verify the layout itself. Finally, it
+makes a terminating delegation of all packages ("`*`") to the
+“package-and-in-toto-metadata-signer'' role. It MUST use a different set
```
JustinCappos: Perhaps explain why it is important to have separate keys from the root.

ITE/2/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+      }
+    }
+    ...,
+
+  }
+}
+----
+
+The snapshot and timestamp roles SHOULD use (1, 1) online keys, and
+their metadata SHOULD expire in 1 day. The in-toto "`packager`"
```
JustinCappos: How does one decide which expiry makes sense?

ITE/2/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+functionary, as well as the TUF timestamp, snapshot, and
+"`package-and-in-toto-metadata-signer`" roles, MAY share the same (1, 1)
+online keys. Furthermore, these online TUF roles MAY share the same
+metadata expiration period (e.g., 1 day). Online keys are used so that
+robots can use the key to sign new metadata on demand. When the in-toto
+or TUF implementation supports it, it is RECOMMENDED that online keys
+be protected from  exfiltration using a Hardware Security Module
+(HSM).
```
JustinCappos: What advantages does an HSM give?  Talking about the real world impact of this is helpful.

ITE/2/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+[arabic]
+. Compromise any part of the infrastructure between, but not including,
+developers and end-users. Thus, attackers can tamper with any files and
+keys stored on the infrastructure, including those situations where the
+key itself is unknown (e.g., due to hardware protection), but where the
+attacker is nevertheless able to sign malicious packages using the key.
+. Respond to user requests. This can be done either by acting as a
+man-in-the-middle, or compromising the infrastructure as above.
+
+We assume that attackers cannot access or reverse-engineer offline keys,
```
JustinCappos: Thresholding is worth mentioning here.

ITE/2/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+The changes discussed here require adding _at least_ the following
+infrastructure and resources:
+
+* *An offline key ceremony to generate and rotate all private and public
+keys for TUF and in-toto.* Operations security (OPSEC) is extremely
+important here into ensuring that offline private keys are not leaked or
+exported by accident. Enough backups should also be made to ensure that
+the keys for the entire system can be rotated when necessary, such as
+removing or adding developer keys, which is an additional cost.
+* *Provisioning developers with signing keys (possibly with GPG on
```
JustinCappos: talk about partial adoption somewhere?

ITE/2/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+preserved for package managers that have not yet been updated to use
+TUF.* These package managers should be updated to newer versions which
+use both TUF and in-toto.
+* *When rotating TUF root keys, all versions of the root TUF metadata
+from version 1 to N, where N is the latest version, should be preserved
+so that package managers can update from one version to the next.* This
+assumes that any version N+1 is signed by at least a threshold of root
+keys listed in version N.
+* *Package managers should use
+https://github.com/theupdateframework/taps/blob/master/tap6.md[TAP **6]*
```
JustinCappos: More is needed here from the TUF side (not you in this doc).  Marina is working on this.  Would love to have your eyeballs when ready.

trishankatdatadog:
@SantiagoTorres I need to write the CNAB TUF-in-toto standard. Could you please help address Justin's comments? Thanks!

trishankatdatadog:
Thanks for all your help, @adityasaky!

@JustinCappos, could we please get another review?

ITE/2/README.adoc
```diff
@@ -0,0 +1,532 @@
...
+}
+----
+
+The targets role signs targets metadata about the in-toto root layout,
+and all of the public keys used to verify the layout itself. Finally, it
+makes a terminating delegation of all packages ("`*`") to the
+“package-and-in-toto-metadata-signer'' role. It MUST use a different set
+of (m, n) offline keys from the root role (to ensure that a compromise of
+these keys do not lead to a compromise of the root role as well), where
+it is RECOMMENDED that n >= 2, and m >= ceiling(n / 2). Finally, its
```
JustinCappos, line 245: Can you give more context about these choices?  The reason why you choose these may be very interesting to those wanting to customize your advice for their context.  It's a little bit in between a spec and a best practices guide as it is, so I'm having a hard time wrapping my head around parts of it...

ITE/2/README.adoc
```diff
@@ -0,0 +1,532 @@
...
+so that package managers can update from one version to the next.* This
+assumes that any version N+1 is signed by at least a threshold of root
+keys listed in version N.
+* *Package managers should use
+https://github.com/theupdateframework/taps/blob/master/tap6.md[TAP **6]*
+to determine how to respond when a TUF repository has significantly
+updated its specification version.**
+* *Neither a TUF nor in-toto implementation should make
+backward-incompatible changes to how signatures are generated, so that
+previous package managers are able to continue to install new packages.*
```
JustinCappos, line 498: This may be tricky as support for crypto agility is added.  In the future, there is hope that significant future-proofing properties can be maintained.

SantiagoTorres:
Something that I would like this PR to specify is what is the key for the custom in-toto metadata. I think the DD implementation uses simply `in-toto`. I was thinking of following something akin to what SMTP does and use `x-in-toto` for the key.

SantiagoTorres:
Another aside that I'm thinking of is if we want to "abstract out" the roles suggested to not mandate a pipeline, but rather the minimum duties provided:

1. We have a role collecting evidence (i.e., the functionaries)
2. We *also* have a role that's taking the delivered product and adding targets metainformation to them.

My rationale for this is to not say "you need to have a packager" but rather "somebody that's wrapping everything together (which btw it fits very well with the packager, look at ITE-3)"

Thoughts?

trishankatdatadog:
Okay, I think we can make these changes. Let me talk to @adityasaky and get back to you...

trishankatdatadog:
Ok, @adityasaky and I made some changes! Are we good now?

SantiagoTorres:
can we resolve the conflicts?

trishankatdatadog:
Done!

SantiagoTorres:
Thanks! 

I was reading and I wonder if we'd like to use the tuf in-toto-demo (with changes) as the reference impl for this. What do you think?

trishankatdatadog:
> I was reading and I wonder if we'd like to use the tuf in-toto-demo (with changes) as the reference impl for this. What do you think?

Agreed, and updated link

trishankatdatadog:
Hi, any news on this, pls?

SantiagoTorres:
Oh, sorry I thought I merged. Could you also go ahead and announce it on the ML so we can start gathering feedback?

PR #6, opened 2019-10-08
ITE-4: Add generic URI schemes for artifacts in in-toto metadata

adityasaky:
This ITE proposes allowing generic URI schemes to refer to abstract entities in in-toto metadata.

### DIFF
--- a/ITE/4/README.adoc
+++ b/ITE/4/README.adoc
@@ -504,8 +504,8 @@ abstract resources securely. It is quite possible that the contents of these
 resources need to be serialized in some manner that allows for their hashing,
 and these operations when performed unsafely can lead to severe vulnerabilities
 if the resource is controlled by a malicious actor. Implementors must also take
-care to ensure they handle situations where a resource may be unavailable, and
-fail appropriately.
+care to ensure ITE-4 compliant systems are capable of handling situations where
+an abstract resource is unavailable, and fail appropriately.
 
 NOTE: There is an ongoing discussion about limiting the scope of this ITE to
 static or unchanging resources. Thread: https://github.com/in-toto/ITE/issues/7
--- a/ITE/4/README.adoc
+++ b/ITE/4/README.adoc
@@ -188,7 +188,7 @@ GitHub has more abstract entities such as Pull Requests and Issues. These
 entities can be referred to directly using the URI schemes proposed in this ITE
 and help provide attestations about these artifacts. Consider:
 
-===== Attestations for merging a pull request into `master`
+===== in-toto link attestations for merging a pull request into `master`
 
 A pull request is a proposal to make changes to a repository. Changes are either
 made on a separate branch on the same repository or a branch on a fork of the
@@ -228,7 +228,7 @@ commit. The attestation could look something like:
 This step is accepting the pull request as a material and is recording the merge
 commit as a product.
 
-===== Attestations for GitHub Actions building from a merge commit
+===== in-toto link attesations for GitHub Actions building from a merge commit
 
 GitHub Actions can be used to set up a workflow for continuous integration (CI).
 Workflows can be triggered on push and an attestation can be generated for the
--- a/ITE/4/README.adoc
+++ b/ITE/4/README.adoc
@@ -182,6 +182,38 @@ a software package in detail. They're composed of several entities that have
 unique identifiers. These identifiers can be used in in-toto metadata embedded
 in SPDX documents to refer to the respective entities.
 
+===== in-toto link attestation for packaging SPDX files into an SPDX package
+
+An in-toto attestation can be bundled into an SPDX document to show the chain of
+custody for the elements the SPDX document refers to. Here, we see how
+provenance can be attested to for File and Package entities in an SPDX document.
+
+```
+{
+    "_type": "link",
+    "name": "package-ghostscript-9.21.tar.gz",
+    "command": "<COMMAND>",
+    "materials": {
+        "spdx:SPDXRef-141-File-83pv-RKSJ-H-d51620a4d7d9aeca3a1cbe5ef201513f98d65f98": <HASH>,
+        "spdx:SPDXRef-271-File-AUTHORS.md-109c93392646b4d55e3ca62c5b578a9ac7cc159f": <HASH>,
+        "...": "..."
+    },
+    "products": {
+        "spdx:SPDXRef-Pkg-ghostscript-9.21.tar.gz-6f60d7fcb5eef6a8bec5abedf21c6a7008a8c0c7": <HASH>
+    },
+    "byproducts": {
+        "stderr": "",
+        "stdout": "",
+        "return-value": ""
+    },
+    "environment": {
+        "variables": "",
+        "filesystem": "",
+        "workdir": ""
+    }
+}
+```
+
 ==== Use Case 2: GitHub Entities
 
 GitHub has more abstract entities such as Pull Requests and Issues. These
--- a/ITE/4/README.adoc
+++ b/ITE/4/README.adoc
@@ -205,13 +205,19 @@ enables them to be reused by different stakeholders looking to hash the same
 abstract entities.
 
 ```
-define hash_artifacts(string generic_uri, list hash_algorithms, kwargs) \
-    returns map artifact_hash_map
+function hash_artifacts(string generic_uri, list hash_algorithms, kwargs) \
+        returns map artifact_hash_map
+    contained_uris <- resolve_uri(generic_uri)
+    for each uri in contained_uris do
+        hashable_representation <- get_hashable_representation
+        set artifact_hash_map[uri] <- hash(hashable_representation)
 
-define resolve_uri(string generic_uri, kwargs) returns list contained_uris
 
-define get_hashable_representation(string generic_uri, kwargs) returns bytes \
-    hashable_representation
+function resolve_uri(string generic_uri, kwargs) returns list contained_uris
+
+
+function get_hashable_representation(string generic_uri, kwargs) returns bytes \
+        hashable_representation
 ```
 
 `hash_artifacts` closely mirrors the general workflow of hashing artifacts in
@@ -240,9 +246,16 @@ It is quite likely that a generic URI specified may point to a collection of
 abstract entities. In the present framework where all artifacts are part of the
 filesystem, this is akin to using the path of a directory to record all the
 contents of the directory as the actual artifacts of the step. Therefore,
-`resolve_uri` is used to resolve such URIs to multiple different artifacts that
-they contain. Each of them must be hashed using their corresponding hashable
-representations.
+in these instances, a method `resolve_uri` must be used to resolve the
+collective URI into the URIs of many different artifacts that are contained or
+represented by the collective URI. Each of them must be hashed using their
+corresponding hashable representations.
+
+Such a method is optional and the operations performed to unpack collective URIs
+into individual URIs need not be implemented if the context or type or abstract
+resources do not require them. All the methods defined above can contain
+additional arguments that may be used to control these and other options of
+operation.
 
 [[motivation]]
 == Motivation
--- a/ITE/4/README.adoc
+++ b/ITE/4/README.adoc
@@ -379,17 +379,48 @@ information for.
 [[security]]
 == Security
 
-It's important to note that file paths are a form of URIs too, and can be
-represented with the `scheme` `file://`. We have also discovered that `MATCH`
-and other artifact rules use only the URI strings themselves to find the
-appropriate entries in the link metadata. The changes proposed in this ITE,
-therefore, have no effect on the security guarantees made by the in-toto
-specification.
+As per the in-toto specification, the only direct interaction in-toto tools have
+with artifacts is to record their hashes using one or more cryptographic hash
+algorithms. Other artifact operations such as verifying the artifact rules
+rely on the hashes recorded in the link metadata.
 
-Implementing systems that conform to this ITE must, however, be careful with how
-they compute cryptographic hashes of abstract entities. The implementors must
-understand the abstract entities they are hashing so as to be able to do so
-consistently.
+We found that the extending the recording of hashes to abstract resources, some
+of which may live at remote locations, is more complicated than the recording of
+hashes of artifacts in the local filesystem. It is, therefore, important for
+ITE-4 compliant in-toto tools to handle the resolving of generic URIs to these
+abstract resources securely. It is quite possible that the contents of these
+resources need to be serialized in some manner that allows for their hashing,
+and these operations when performed unsafely can lead to severe vulnerabilities
+if the resource is controlled by a malicious actor.
+
+Further, it is likely that abstract resources change more frequently, both in
+content and format, and implementors must take care to identify how these
+contents are recorded, as well what specific information is recorded for a
+particular type of entity. Otherwise, the in-toto verification workflow can be
+plagued by failures due to the lack of availability of the artifact as
+previously recorded. For example, implementors who decide how to resolve GitHub
+pull request must decide what information encoded in a pull request must be
+hashed. If they are considering the comments of various users when hashing the
+pull request, this can lead to differences in the hashes recorded by different
+steps in the supply chain even if the changes proposed in the pull request are
+the same. This will of course eventually lead to a failure of the in-toto
+verification workflow.
+
+As always, it is also necessary to consider the actors who can make changes to
+an abstract entity. This is perhaps slightly exacerbated in the case of abstract
+resources as the content and format being hashed aren't as specific as artifacts
+in the local filesystem. In the above example, for a public repository, *any*
+GitHub user can comment on a pull request, so for an implementation that also
+considers comments when recording the hash, any user can potentially cause a
+failure of the verification workflow. This can potentially be leveraged by a
+malicious actor to target automated pipelines that rely on in-toto verification.
+
+Finally, our analysis showed that verification of artifact rules specified in
+in-toto layouts use the hashes recorded while generating the link metadata and
+do not record afresh any of the hashes. So the security properties of artifact
+rules verification rely on the above points - ensuring the hashes are recorded
+correctly for a given context. The change in the URI format has no direct impact
+on the artifact rules and their verification.
 
 [[infrastructure-requirements]]
 == Infrastructure Requirements
--- a/ITE/4/README.adoc
+++ b/ITE/4/README.adoc
@@ -146,10 +146,12 @@ tokens that all conforming implementations MUST support. Instead, it merely
 specifies what any generic URI must look like, and what rules the system MUST
 follow.
 
+==== Artifact Rules
+
 in-toto artifact rules are a means to specify what artifacts the project owner
 expects in every step of the supply chain in the in-toto layout. in-toto
-provides the following rules that together attempt to describe the different
-types of operations or actions that supply chain functionaries carry out.
+provides the certain rules that together attempt to describe the different types
+of operations or actions that supply chain functionaries carry out.
 
 *MATCH Rule*
 
@@ -180,6 +182,13 @@ different checks by verifying if artifacts matching the pattern occur or do not
 occur in the materials or products sections of the relevant link. This does not
 involve resolving the pattern or URI into the artifact itself, and so this ITE
 does not affect the working of these rules.
+
+However, it is quite likely that the contents referred to using generic URIs
+change more frequently than the traditional artifacts in the filesystem. The
+expectation is that either these changes are recorded with link attestations of
+their own, making them part of the supply chain, or metadata is replaced by
+authorized functionaries. The security implications of frequently changing
+generic resources are discussed in the Security section below.
 
 [[motivation]]
 == Motivation
--- a/ITE/4/README.adoc
+++ b/ITE/4/README.adoc
@@ -596,10 +596,9 @@ maintain that consistency throughout, ensuring a single root is maintained.
 Similarly, it is possible for generic URIs that point to remote locations to be
 redirected at times. In some cases, the resources are moved to a new location.
 For example, when a user changes their GitHub username, a redirect is setup for
-references to the old username. Similarly, when repositories are renamed, a
-redirect is setup to the new location. Implementors must be careful with
-redirects and must decide based on context and security considerations if the
-compliant system should follow them to the destination or not.
+references to the old username. Implementors must be careful with redirects and
+must decide based on context and security considerations if the compliant system
+should follow them to the destination or not.
 
 ==== Dynamic data in remote abstract entities
 
--- a/ITE/4/README.adoc
+++ b/ITE/4/README.adoc
@@ -228,15 +228,18 @@ libraries.
 If in-toto metadata is generated using an implementation of in-toto conforming
 to this ITE, verification using a non-conforming implementation will likely
 fail. It is still possible for verification to continue if the inspections don't
-use generic URIs.
+use generic URIs. This is because the layout may contain inspections that use
+generic URIs.
 
 However, a conforming implementation should be capable of verifying in-toto
-metadata generated using a non-conforming implementation, as a conforming
-system must also conform to the main in-toto specification.
+metadata generated using a non-conforming implementation, as an ITE-4 conforming
+system must also conform to the actual in-toto specification.
 
 It's also possible two conforming systems are unable to verify the other's
 in-toto metadata as they're unaware of how to resolve certain URI tokens used by
-the other.
+the other. This is again because of the possibility of inspections containing
+URIs that the other system is unable to resolve and calculate cryptographic
+information for.
 
 [[security]]
 == Security
--- a/ITE/4/README.adoc
+++ b/ITE/4/README.adoc
@@ -56,7 +56,7 @@ in-toto link metadata files currently look like this:
 ```
 {
     "_type": "link",
-    "_name": "<NAME>",
+    "name": "<NAME>",
     "command": "<COMMAND>",
     "materials": {
         "<PATH>": <HASH>,
@@ -92,7 +92,7 @@ This ITE proposes the following structure for the URI.
 `<scheme>:<hier-part>`
 
 This structure is derived from RFC 3986. The `scheme` defines the token or
-identifier for the type of resource the URI refers to . It can also contain
+identifier for the type of resource the URI refers to. It can also contain
 other relevant information that indicates how the resource can be accessed. The
 `hier-part` identifies the path or location of the resource.
 
@@ -134,7 +134,7 @@ documents, the URI can also be used to refer to other entities elsewhere in the
 document.
 
 Artifact Rules are a means to specify what artifacts the project owner expects
-in every step of the supply chain in the in-toto layout. in-toto povides the
+in every step of the supply chain in the in-toto layout. in-toto provides the
 following rules that together attempt to describe the different types of
 operations or actions that supply chain functionaries carry out.
 
@@ -245,7 +245,7 @@ It's important to note that filepaths are a form of URIs too, and can be
 represented with the `scheme` `file://`. We have also discovered that `MATCH`
 and other artifact rules use only the URI strings themselves to find the
 appropriate entries in the link metadata. The changes proposed in this ITE,
-therefore, have no effect on the security guarantees made by th in-toto
+therefore, have no effect on the security guarantees made by the in-toto
 specification.
 
 Implementing systems that conform to this ITE must, however, be careful with how
--- a/ITE/4/README.adoc
+++ b/ITE/4/README.adoc
@@ -38,14 +38,14 @@ endif::[]
 
 This ITE proposes a generic Unique Resource Identifier (URI) scheme for in-toto
 artifacts. Currently, the in-toto specification only allows for files in the
-local filesystem to be the artifacts of link metadata. This ITE also defines the
-scheme for a generic URI and provides a behaviour of in-toto implementations
-that support generic URIs.
+local filesystem to be the artifacts of link metadata. This ITE defines the
+scheme for a generic URI and provides a description of the expected behaviour of
+in-toto implementations that support generic URIs.
 
 [[specification]]
 == Specification
 
-The current in-toto specification must be modified to provide support for the
+The current in-toto specification must be modified to provide support for
 generic URI schemes. The current specification allows only for files in the
 local filesystem to be set as artifacts. With the addition of other entities,
 the parts of the specification that directly interact or handle the artifacts
@@ -114,8 +114,8 @@ This can also instead use GitHub's API specific URI to refer to the same data.
 
 Both the above URIs refer to the same abstract entity - pull request 250 to
 in-toto's reference implementation. However, taken directly, they correspond to
-different data. Therefore, the implementing system must resolve the URI and use
-it consistently.
+different data. Therefore, the implementing system must resolve the URI and thus
+the artifact in a consistent manner.
 
 An example of the product field for link metadata with a generic URI is as
 follows:
@@ -354,7 +354,7 @@ use generic URIs.
 
 However, a conforming implementation should be capable of verifying in-toto
 metadata generated using a non-conforming implementation, as an ITE-4 conforming
-system must also conform to the actual in-toto specification.
+system MUST also conform to the actual in-toto specification.
 
 It's also possible two conforming systems may be unable to verify the other's
 in-toto metadata as they're unaware of how to resolve certain URI tokens used by
@@ -402,4 +402,3 @@ This ITE currently proposes no prototypes.
 
 * link:https://tools.ietf.org/html/rfc3986[Uniform Resource Identifier (URI): Generic Syntax]
 * link:https://github.com/in-toto/docs/blob/master/in-toto-spec.md[in-toto Specification]
-* link:https://pypi.org/project/rfc3986/[RFC 3986 library for Python]
--- a/ITE/4/README.adoc
+++ b/ITE/4/README.adoc
@@ -42,6 +42,26 @@ local filesystem to be the artifacts of link metadata. This ITE also defines the
 scheme for a generic URI and provides a behaviour of in-toto implementations
 that support generic URIs.
 
+[[motivation]]
+== Motivation
+
+ITE 4 is motivated by the following use cases.
+
+==== Use Case 1: SPDX Document Identifiers
+
+The Software Package Data Exchange (SPDX) is an open standard for communicating
+software bill of material information such as components, licenses, copyrights,
+and security references. Each document is a comprehensive report that describes
+a software package in detail. They're composed of several entities that have
+unique identifiers. These identifiers can be used in in-toto metadata embedded
+in SPDX documents to refer to the respective entities.
+
+==== Use Case 2: GitHub Entities
+
+GitHub has more abstract entities such as Pull Requests and Issues. These
+entities can be referred to directly using the URI schemes proposed in this ITE
+and help provide attestations about this data.
+
 [[specification]]
 == Specification
 
@@ -83,9 +103,8 @@ The `materials` and `products` fields contain key-value pairs. The URI for each
 artifact is its path in the filesystem. With this ITE active, the artifact could
 instead have a generic URI which can be resolved to some external resource that
 is part of the supply chain. The URI MUST have a fixed structure and MUST
-contain all the information necessary for any verifier to obtain the resource it
-refers to. Therefore, any verifier MUST be able to take the URI in isolation and
-know how and where to find the artifact.
+contain all the information necessary for the verifier to resolve if any part of
+the verification workflow requires it.
 
 This ITE proposes the following structure for the URI.
 
@@ -112,10 +131,6 @@ This can also instead use GitHub's API specific URI to refer to the same data.
 
 `github+https://api.github.com/repos/in-toto/in-toto/pulls/250`
 
-The in-toto system generating link metadata for this artifact MUST be capable
-of recognizing that it possesses a generic URI, and the system MUST be able to
-resolve this resource and compute its cryptographic hash information. 
-
 An example of the product field for link metadata with a generic URI is as
 follows:
 
@@ -131,12 +146,19 @@ follows:
 Similarly, an SPDX URI can be used to directly fetch the corresponding documents
 from a specified location. For in-toto metadata that exist within SPDX
 documents, the URI can also be used to refer to other entities elsewhere in the
-document.
+document. Resolving such URIs that point to local entities elsewhere in the SPDX
+document is an implementation level detail - but it's clear that the URI
+contains all the information necessary to identify the entity in that context.
 
-Artifact Rules are a means to specify what artifacts the project owner expects
-in every step of the supply chain in the in-toto layout. in-toto provides the
-following rules that together attempt to describe the different types of
-operations or actions that supply chain functionaries carry out.
+It is important to note that this ITE doesn't mandate any specific generic URI
+tokens that all conforming implementations MUST support. Instead, it merely
+specifies what any generic URI must look like, and what rules the system MUST
+follow.
+
+in-toto artifact rules are a means to specify what artifacts the project owner
+expects in every step of the supply chain in the in-toto layout. in-toto
+provides the following rules that together attempt to describe the different
+types of operations or actions that supply chain functionaries carry out.
 
 ==== `MATCH <pattern> [IN <source-path-prefix>] WITH (MATERIALS|PRODUCTS) [IN <destination-path-prefix>] FROM <step>`
 
@@ -163,26 +185,6 @@ in-toto provides several other artifact rules - specifically `ALLOW`,
 check that the provided pattern exists are does not exist in the relevant link
 metadata, without ever resolving the pattern into the artifact itself.
 Therefore, this ITE does not affect the working of these rules.
-
-[[motivation]]
-== Motivation
-
-ITE 4 is motivated by the following use cases.
-
-==== Use Case 1: SPDX Document Identifiers
-
-The Software Package Data Exchange (SPDX) is an open standard for communicating
-software bill of material information such as components, licenses, copyrights,
-and security references. Each document is a comprehensive report that describes
-a software package in detail. They're composed of several entities that have
-unique identifiers. These identifiers can be used in in-toto metadata embedded
-in SPDX documents to refer to the respective entities.
-
-==== Use Case 2: GitHub Entities
-
-GitHub has more abstract entities such as Pull Requests and Issues. These
-entities can be referred to directly using the URI schemes proposed in this ITE
-and help provide attestations about this data.
 
 [[reasoning]]
 == Reasoning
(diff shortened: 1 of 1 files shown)


ITE/4/README.adoc
```diff
@@ -0,0 +1,283 @@
...
+        "sha256": "7ee0617..."
+    },
+    "...": ...
+}
+```
+
+Similarly, an SPDX URI can be used to directly fetch the corresponding documents
+from a specified location. For in-toto metadata that exist within SPDX
+documents, the URI can also be used to refer to other entities elsewhere in the
+document.
```
lukpueh: Is the reader of this ITE supposed to be familiar with SPDX and/or with in-toto/SPDX integration efforts? If not, some elaboration might be in order.

Why is an SPDX URI different form any other URI? And why is an SPDX URI that points to a resources in the same document different from any other SPDX URI?

ITE/4/README.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+The Require rule specifies that an artifact matched by this rule MUST appear as
+a material or product of the particular step. This rule is not affected by this
+ITE.
+
+=== `MODIFY <pattern>`
+
+The Modify rule specifies that an artifact matched by this rule MUST appear as
+both a material and product of the step, and the hash of the artifact MUST
+differ indicating the execution of the step changed the artifact.
```
lukpueh: Is it necessary to mention all rules here to this extent?

To me it makes sense to clarify that processing rules does not require resolving any paths, with or without this ITE. You already say this quite aptly above in the match rule section, i.e. *"...does not entail resolving the URIs in any form..."*. 

I suggest to, instead of iterating over all rules here, you only leave that one paragraph that briefly explains what artifact rules are good for, and then explain how this ITE does not affect any of the rule's functioning.

The  `"MATCH pull/250 ..."` example, on the other hand, could use some explanation about the context. What kind of supply chain are we talking about here. What does the `"merge-go-repository"` step do? Why would the exact same GitHub PR page exist on two different repos?

ITE/4/README.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+[[motivation]]
+== Motivation
+
+ITE 4 is motivated by the following use cases.
+
+<not just licenses, all spdx materials and "refer" to those instances>
+
+<OCI, Datadog, save space by common references to file hashes>
```
lukpueh: Are these, i.e. *"<...>"*, leftover todo notes?

ITE/4/README.adoc
```diff
@@ -0,0 +1,283 @@
...
+fail. It is still possible for verification to continue if the inspections don't
+use generic URIs.
+
+However, a conforming implementation should be capable of verifying in-toto
+metadata generated using a non-conforming implementation, as a conforming
+system must also conform to the main in-toto specification.
+
+It's also possible two conforming systems are unable to verify the other's
+in-toto metadata as they're unaware of how to resolve certain URI tokens used by
+the other.
```
lukpueh: Isn't this ITE about resolving URIs and hashing the contents of the documents they point to, when **generating** link metadata (e.g. in `in-toto-run`)?

How is verification (e.g. through `in-toto-verify`) affected by this ITE? As you say above, at the time of verification no more URIs are resolved.

The only compatibility issue I can see on the verification side is the link metadata schema. But from quickly looking at the specs and the reference implementation, artifact paths already don't have any restrictions that wouldn't also allow for a "`<scheme>:<hier-part>`"-string.

ITE/4/README.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+[[security]]
+== Security
+
+It's important to note that filepaths are a form of URIs too, and can be
+represented with the `scheme` `file://`. We have also discovered that `MATCH`
+and other artifact rules use only the URI strings themselves to find the
+appropriate entries in the link metadata. The changes proposed in this ITE,
+therefore, have no effect on the security guarantees made by th in-toto
```
lukpueh:
```suggestion
therefore, have no effect on the security guarantees made by the in-toto
```

ITE/4/README.adoc
adityasaky: Yikes, good catch. Thanks!

ITE/4/README.adoc
```diff
@@ -0,0 +1,283 @@
...
+local filesystem to be set as artifacts. With the addition of other entities,
+the parts of the specification that directly interact or handle the artifacts
+must be updated.
+
+in-toto link metadata files currently look like this:
+
+```
+{
+    "_type": "link",
+    "_name": "<NAME>",
```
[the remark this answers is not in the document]
adityasaky: That's a good point, I'll open an issue later. I copied this segment straight from the specification, and didn't notice.

I think `name` is pretty much established at this point, @SantiagoTorres?

ITE/4/README.adoc
adityasaky: > I suggest to, instead of iterating over all rules here, you only leave that one paragraph that briefly explains what artifact rules are good for, and then explain how this ITE does not affect any of the rule's functioning.

That does make sense. The MATCH rule is the only real confusing one so I think it's better to explore that a bit while just explaining the rest briefly.

> The `"MATCH pull/250 ..."` example, on the other hand, could use some explanation about the context. What kind of supply chain are we talking about here. What does the "merge-go-repository" step do? Why would the exact same GitHub PR page exist on two different repos?

That's a good point, and I'll work on an improved scenario. We're also not talking about just the pages as much as some more abstract representation of what a pull request is, but I think the point you made still stands. Thanks!

ITE/4/README.adoc
adityasaky: There's a bit about SPDX in motivation, and I'll expand that perhaps. As far as "different from any other URI", no, I don't see it being so. Let me work on the phrasing here.

ITE/4/README.adoc
lukpueh: Oh right, inspections. I think it's worth being explicit there.

iamwillbar:
At a high-level this looks reasonable to me, a couple of concrete examples that tie together in a supply chain would probably help though.

ITE/4/README.adoc
```diff
@@ -0,0 +1,361 @@
...
+artifacts. It's important to note that the contents of the resource are hashed,
+as opposed to the URI itself.
+
+To that end, it's important for a URI scheme to have a single mapping for the
+service it refers to. Consider an example of a GitHub specific resource - a pull
+request or an issue.
+
+`github+https://github.com/in-toto/in-toto/pull/250`
+
+This can also instead use GitHub's API specific URI to refer to the same data.
```
lukpueh: What does "same data" mean? Speaking in abstract terms `github+https://github.com/in-toto/in-toto/pull/250` and `github+https://api.github.com/repos/in-toto/in-toto/pulls/250` point to the same *thing*. But how do those *things* get hashed?
```
`curl https://github.com/in-toto/in-toto/pull/250 | shasum -a 256` != `curl https://api.github.com/repos/in-toto/in-toto/pulls/250 | shasum -a 256`
```

ITE/4/README.adoc
```diff
@@ -0,0 +1,361 @@
...
+```
+
+This step is accepting the pull request as a material and is recording the merge
+commit as a product.
+
+===== in-toto link attesations for GitHub Actions building from a merge commit
+
+GitHub Actions can be used to set up a workflow for continuous integration (CI).
+Workflows can be triggered on push and an attestation can be generated for the
+build that as a result.
```
lukpueh: I think this sentence is incomplete.

ITE/4/README.adoc
```diff
@@ -0,0 +1,361 @@
...
+        "filesystem": "",
+        "workdir": ""
+    }
+}
+```
+
+This step is accepting the pull request as a material and is recording the merge
+commit as a product.
+
+===== in-toto link attesations for GitHub Actions building from a merge commit
```
lukpueh:
```suggestion
===== in-toto link attestation for GitHub Actions building from a merge commit
```

ITE/4/README.adoc
```diff
@@ -0,0 +1,361 @@
...
+    }
+}
+```
+
+[[reasoning]]
+== Reasoning
+
+This ITE proposes a change in the URI scheme of artifacts in in-toto metadata.
+The proposed change closely matches the URI structure laid out in RFC 3986. The
+URI structure propsed in that document is widely accepted and is versatile
```
lukpueh:
```suggestion
URI structure proposed in that document is widely accepted and is versatile
```

ITE/4/README.adoc
```diff
@@ -0,0 +1,361 @@
...
+
+By default, the commits that make up the change are integrated into the target
+branch and an additional merge commit is created to indicate the act of merging.
+It's also possible to merge a pull request without creating a separate merge
+commit. The attestation could look something like:
+
+```
+{
+    "_type": "link",
+    "name": "merge-pull-request-250",
```
lukpueh: Does this mean that there will be a layout that has a `merge-pull-request-250` step? Or am I thinking too far ahead?

ITE/4/README.adoc
```diff
@@ -0,0 +1,361 @@
...
+
+It's important to note that filepaths are a form of URIs too, and can be
+represented with the `scheme` `file://`. We have also discovered that `MATCH`
+and other artifact rules use only the URI strings themselves to find the
+appropriate entries in the link metadata. The changes proposed in this ITE,
+therefore, have no effect on the security guarantees made by the in-toto
+specification.
+
+Implementing systems that conform to this ITE must, however, be careful with how
+they compute cryptographic hashes of abstract entities.
```
lukpueh: Hashing the contents of ITE-4-addressable artifacts is not in the scope of ITE-4?

ITE/4/README.adoc
```diff
@@ -0,0 +1,361 @@
...
+The current verification workflow for the `MATCH` rule compares the
+cryptographic hashes of the relevant artifacts from the respective link files.
+This is a straightforward comparison and does not entail resolving the URIs in
+any form. The `IN` clauses which are used to specify path prefixes are
+subtracted from the URI strings. Therefore, we conclude that the functioning of
+the rule doesn't change with this ITE. An example `MATCH` rule with generic URIs
+is as follows:
+
+`MATCH pull/250 IN github+https://github.com/in-toto/in-toto/ WITH PRODUCTS IN
+github+https://github.com/in-toto/in-toto-golang/pull/ FROM merge-go-repository`
```
lukpueh: I suggest to update the MATCH rule example to relate to the use cases below.

ITE/4/README.adoc
adityasaky: Whoops, thanks!

ITE/4/README.adoc
```diff
@@ -0,0 +1,361 @@
...
+
+*Other artifact rules*
+
+in-toto provides several other artifact rules - specifically `ALLOW`,
+`DISALLOW`, `CREATE`, `DELETE`, `REQUIRE`, and `MODIFY`. All these rules merely
+check that the provided pattern exists are does not exist in the relevant link
```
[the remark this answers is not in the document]
adityasaky: Actually, I realized I'm not happy with this description, and I think I'm going to rewrite it slightly. Thanks for catching the `s/are/or` though.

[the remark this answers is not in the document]
adityasaky: What do you think of 96049bc? Less of a rewrite than I thought.

ITE/4/README.adoc
```diff
@@ -0,0 +1,404 @@
...
+
+[[motivation]]
+== Motivation
+
+ITE 4 is motivated by the following use cases.
+
+==== Use Case 1: SPDX Document Identifiers
+
+The Software Package Data Exchange (SPDX) is an open standard for communicating
+software bill of material information such as components, licenses, copyrights,
```
trishankatdatadog:
```suggestion
software bill of material (SBoM) information such as components, licenses, copyrights,
```

ITE/4/README.adoc
```diff
@@ -0,0 +1,404 @@
...
+instead have a generic URI which can be resolved to some external resource that
+is part of the supply chain. The URI MUST have a fixed structure and MUST
+contain all the information necessary for the verifier to resolve if any part of
+the verification workflow requires it.
+
+This ITE proposes the following structure for the URI.
+
+`<scheme>:<hier-part>`
+
+This structure is derived from RFC 3986. The `scheme` defines the token or
```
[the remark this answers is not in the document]
adityasaky: It's in the references. Should a link go here too?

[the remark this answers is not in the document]
trishankatdatadog: Yes, please

ITE/4/README.adoc
adityasaky: Added link to CD Foundation's SIG-SBoM, NTIA's Software Component Transparency, and CISQ/OMG's SBoM working group. The NIST stuff seemed to overlap with some of the NTIA material. Thoughts?

adityasaky:
Thanks for the review @trishankatdatadog!

> I understand why, but I think the ITE punts too much of the actual implementation to the runtime. There shouldn't be two different hashes for two files, no? But it's conceivable that two different implementations can get two different hashes for the same external resource.

Yes, it's possible, and is part of the compatibility analysis. I'm not sure we can lay down the behaviour of how specific tokens are resolved and the content is hashed in a general manner, but we did consider perhaps a registry of tokens with defined behaviours. However, that'll have to be further down the line and out of scope of this ITE. Do you have other ideas on how we can generalise this?

> Should discuss fixing sources of nondeterminism, which are notorious on web resources.

While I think I understand what you mean, do you have some examples I can look at and analyse?

ITE/4/README.adoc
```diff
@@ -0,0 +1,462 @@
...
+
+- that all the artifact rules behave as described in the specification
+- that cryptographic hashes of data in abstract entities change with changes in
+in the data - in effect, this would test how the data is transformed into a form
+that can be hashed
+
+[[prototype-implementation]]
+== Prototype Implementation
+
+This ITE currently proposes no prototypes.
```
JustinCappos: I think that this is needed for sure.  I'd like to see how this works in practice and have several implementations independently done of the same URI type to see how similar the hashes, etc. are.

ITE/4/README.adoc
```diff
@@ -0,0 +1,462 @@
...
+
+[[testing]]
+== Testing
+
+In an in-toto system conforming to this ITE, it is important to test:
+
+- that all the artifact rules behave as described in the specification
+- that cryptographic hashes of data in abstract entities change with changes in
+in the data - in effect, this would test how the data is transformed into a form
+that can be hashed
```
JustinCappos: I'm slightly confused here.  (This is likely related to my other points)

ITE/4/README.adoc
```diff
@@ -0,0 +1,462 @@
...
+represented with the `scheme` `file://`. We have also discovered that `MATCH`
+and other artifact rules use only the URI strings themselves to find the
+appropriate entries in the link metadata. The changes proposed in this ITE,
+therefore, have no effect on the security guarantees made by the in-toto
+specification.
+
+Implementing systems that conform to this ITE must, however, be careful with how
+they compute cryptographic hashes of abstract entities. The implementors must
+understand the abstract entities they are hashing so as to be able to do so
+consistently.
```
JustinCappos: And this must be coordinated, and items may change over time, etc...  I feel like there are a lot of really hard problems lurking here.

ITE/4/README.adoc
```diff
@@ -0,0 +1,462 @@
...
+
+[[security]]
+== Security
+
+It's important to note that file paths are a form of URIs too, and can be
+represented with the `scheme` `file://`. We have also discovered that `MATCH`
+and other artifact rules use only the URI strings themselves to find the
+appropriate entries in the link metadata. The changes proposed in this ITE,
+therefore, have no effect on the security guarantees made by the in-toto
+specification.
```
JustinCappos: I need a more comprehensive argument to be convinced.

ITE/4/README.adoc
```diff
@@ -0,0 +1,462 @@
...
+        "filesystem": "",
+        "workdir": ""
+    }
+}
+```
+
+This step is accepting the pull request as a material and is recording the merge
+commit as a product.
+
+===== in-toto link attesation for GitHub Actions building from a merge commit
```
JustinCappos: spelling error **attestation**

ITE/4/README.adoc
```diff
@@ -0,0 +1,462 @@
...
+    "environment": {
+        "variables": "",
+        "filesystem": "",
+        "workdir": ""
+    }
+}
+```
+
+The materials aren't limited to the commits that make up a pull request, but can
+also contain other elements such as reviews or other comments that make up a
```
JustinCappos: If it contains reviews and/or comments which may grow over time, how can you compute a hash over it?

ITE/4/README.adoc
```diff
@@ -0,0 +1,462 @@
...
+discussions with Continuous Delivery (CD) Foundation's
+link:https://github.com/cdfoundation/sig-security-sbom[Special Interest Group for Software Bill of Materials],
+as well as with the National Telecommunication and Information Administration's
+(NTIA) link:https://www.ntia.doc.gov/SoftwareTransparency[Software Component Transparency].
+SPDX and in-toto, along with representatives from NTIA and other stakeholders,
+are also part of the Consortium for Information and Software Quality (CISQ) and
+Object Management Group's (OMG) working group on
+link:https://www.it-cisq.org/software-bill-of-materials/index.htm[Tool-to-tool Software Bill of Materials Exchange].
+
+SPDX documents composed of several entities that have unique identifiers. These
```
JustinCappos: > SPDX documents composed of several entities that have unique identifiers.

missing "are"  (I can't suggest the edit, possibly because I'm typing this while on a flight.)

ITE/4/README.adoc
```diff
@@ -0,0 +1,462 @@
...
+github+https://github.com/in-toto/in-toto FROM merge-pull-request-250`
+
+*Other artifact rules*
+
+in-toto provides several other artifact rules - specifically `ALLOW`,
+`DISALLOW`, `CREATE`, `DELETE`, `REQUIRE`, and `MODIFY`. These rules perform
+different checks by verifying if artifacts matching the pattern occur or do not
+occur in the materials or products sections of the relevant link. This does not
+involve resolving the pattern or URI into the artifact itself, and so this ITE
+does not affect the working of these rules.
```
JustinCappos: I may be a little confused here, but why wouldn't it change these?  If you are creating an item at a URI, why can't in-toto restrict / validate that?

ITE/4/README.adoc
```diff
@@ -0,0 +1,462 @@
...
+The current verification workflow for the `MATCH` rule compares the
+cryptographic hashes of the relevant artifacts from the respective link files.
+This is a straightforward comparison and does not entail resolving the URIs in
+any form. The `IN` clauses which are used to specify path prefixes are
+subtracted from the URI strings. Therefore, we conclude that the functioning of
+the rule doesn't change with this ITE. An example `MATCH` rule with generic URIs
+is as follows:
+
+`MATCH commit/* IN github+https://github.com/in-toto/in-toto/ WITH PRODUCTS IN
+github+https://github.com/in-toto/in-toto FROM merge-pull-request-250`
```
JustinCappos: How do you deal with the fact that the content at a URI may change over time?  Is this intended to be allowed?  Are URIs meant to point to immutable content?

ITE/4/README.adoc
```diff
@@ -0,0 +1,462 @@
...
+
+Similarly, a generic URI can be used to refer to other abstract entities like an
+SPDX element. The URI can be used to directly fetch the corresponding documents
+from a specified location. For in-toto metadata that are embedded within SPDX
+documents, the URI can also be used to refer to other entities elsewhere in the
+document. Resolving such URIs that point to local entities elsewhere in the SPDX
+document is an implementation level detail - but it's clear that the URI must
+contain all the information necessary to identify the entity in that context.
+
+It is important to note that this ITE doesn't mandate any specific generic URI
```
JustinCappos: > It is important to note that this ITE doesn't mandate any specific generic URI
> tokens that all conforming implementations MUST support.

Is this intended to be done in another ITE?  Leaving this unspecified feels like it opens the door for massive amounts of divergence.

ITE/4/README.adoc
```diff
@@ -0,0 +1,462 @@
...
+contain other relevant information that indicates how the resource can be
+accessed. The `hier-part` identifies the path or location of the resource.
+
+A compliant in-toto system must be capable of generating link metadata with
+artifacts that have generic URI references. Therefore, this system must also be
+capable of resolving this URI and calculating cryptographic hashes of these
+artifacts. It's important to note that the contents of the resource are hashed,
+as opposed to the URI itself. This ITE doesn't specify how the contents are
+resolved and hashed by the system generating the metadata for the relevant
+abstract entities. All it specifies is that the resolution and the URI itself
```
JustinCappos: > This ITE doesn't specify how the contents are
> resolved and **hashed** by the system generating the metadata for the relevant
> abstract entities.

If you don't know how to hash something across different implementations, how can you verify items at URIs?  If I implement an in-toto verification routine for my package manager, how do I handle URIs I don't know?  Can different tools that generate in-toto link metadata choose to hash the same URI in different ways?  If so, this is concerning to me.

ITE/4/README.adoc
```diff
@@ -0,0 +1,462 @@
...
+    "environment": {
+        "variables": "<ENV>",
+        "filesystem": "<FS>",
+        "workdir": "<PWD>"
+    }
+}
+```
+
+The `materials` and `products` fields contain key-value pairs. The URI for each
+artifact is its path in the filesystem. With this ITE active, the artifact could
```
JustinCappos: > artifact is its path in the filesystem. With this ITE active, the artifact could
 Is this "could/MAY" or "MUST"?  In other words, are you proposing to change /tmp/buildfile/a.out to be file:///tmp/buildfile/a.out

ITE/4/README.adoc
adityasaky: I think there will be experimental implementations soon because of the ongoing integration with SPDX. Santiago and I will sketch stuff out and I'll update this section with details.

ITE/4/README.adoc
adityasaky: As it stands, an in-toto link records the hashes of artifacts at the time of generation of the link. This doesn't change with this ITE. A link generated for a GitHub PR would capture the information in the PR at that time. If I'm understanding it right, the question then becomes "when is the link generated for a PR", and I suspect that becomes an implementation detail that shouldn't be resolved in this ITE.

ITE/4/README.adoc
adityasaky: Currently, these rules dictate that the pattern specified must appear (or not, depending on the rule) in the step or inspection's link. This is in turn dependent on the in-toto tool generating the metadata being able to record the artifacts correctly, but the verification of these rules ensures the artifact appears in the respective products / materials dictionary. This is not different depending on the type of URI - the pattern must simply appear as a key in the dictionary.

I may be misunderstanding you here, but in-toto's specification for the artifact rules are not dependent on the artifact identifiers being filepaths only, so their restrictions / validation properties should continue to hold.

ITE/4/README.adoc
adityasaky: No, URIs don't point only at immutable content. I think this overlaps with the point above about hashing changing content. The in-toto specification doesn't keep up with changes to artifacts with time. I think the general expectation is that metadata is versioned / re-generated or perhaps the changes are themselves captured with link attestations, but I'd like confirmation. Our analysis was that the temporal aspects of in-toto metadata are not in scope of this ITE.

ITE/4/README.adoc
adityasaky: It can be. We briefly discussed it as an option to establish a registry of URI tokens much further down the line with a separate ITE. Do you think that's feasible?

ITE/4/README.adoc
JustinCappos: > No, URIs don't point only at immutable content. I think this overlaps with the point above about hashing changing content. 

Yes, it does.  Okay, this seems like an important thing to be very clear about early in the document.  What happens if, for example, github changes the format of the data they present at a URI?  Will client verification / signing code need to change?  

> The in-toto specification doesn't keep up with changes to artifacts with time.  I think the general expectation is that metadata is versioned / re-generated or perhaps the changes are themselves captured with link attestations, but I'd like confirmation. 

Can you give an example of what you mean so I can be sure this is clear?

> Our analysis was that the temporal aspects of in-toto metadata are not in scope of this ITE.

These issues become much more important here because the data at URIs is likely to change over time.  I do think you're right that an ITE that covers temporal aspects would help this ITE tremendously.

adityasaky: > Can you give an example of what you mean so I can be sure this is clear?

I mean that if the contents of a file in the filesystem that is recorded as an artifact in in-toto links change, then verification of existing metadata will fail. Either the change itself is recorded as a step and we use its link to validate the change, or the functionary must replace existing metadata with the updated ones, capturing the new hash(es) of the artifact.

> What happens if, for example, github changes the format of the data they present at a URI? Will client verification / signing code need to change?

ITE-4 currently specifies that the data represented as an abstract entity must be resolved and then hashed. So a change in format of the data _could_ require a change in how the URI is resolved to provide hashable data. But this resolving operation should be separate from the verification and signing code. Does that seem sane? It may also just be a difference in semantics, so I suppose the answer is yes, a change of some nature may be necessary.

ITE/4/README.adoc
adityasaky: While it can lead to divergence and problems with cross compatibility (this is explored in the Backward Compatibility section in the context of current implementations being able to verify metadata with generic URIs but perhaps a detailed analysis belongs in the specification section), I'm not sure we can enforce how these resources are hashed, since I believe it's very hard to generalise.

ITE/4/README.adoc
adityasaky: I'll update the phrasing. The idea is for current filesystem paths to co-exist with generic URIs. The filesystem URIs _can_ have `file://` prepended, but it's not "MUST". The important bit here is that if it is a generic URI, it MUST follow the RFC's format.

ITE/4/README.adoc
adityasaky: I'll work on this section.

ITE/4/README.adoc
adityasaky: > Such a generic URI MUST have a fixed structure and MUST contain all the information necessary for the verifier to resolve if any part of the verification workflow requires it.

Does this at 81565d3 clarify?

lukpueh: Minor nit: I would also reword/uppercase ...
```suggestion
artifact is its path in the filesystem. With this ITE active, the artifact CAN
```

I think then it gets even clearer. :)

ITE/4/README.adoc
lukpueh: ITE-4 is probably less specific than it seems through the motivating examples GitHub and SPDX. And I think there is merit in having a generic ITE like this. It wouldn't be feasible to define a generic way of hashing all the resources that might be added through this ITE.

But I also agree with @JustinCappos that it is important that all implementations that support a given scheme should (or rather must) follow the same specification for hashing the resource. @adityasaky, have you thought about how such a specification should be negotiated? Through separate ITEs? Or by the implementers?  Maybe ITE-4 could discuss this?

ITE/4/README.adoc
```diff
@@ -0,0 +1,493 @@
...
+SPDX element. The URI can be used to directly fetch the corresponding documents
+from a specified location. For in-toto metadata that are embedded within SPDX
+documents, the URI can also be used to refer to other entities elsewhere in the
+document. Resolving such URIs that point to local entities elsewhere in the SPDX
+document is an implementation level detail - but it's clear that the URI must
+contain all the information necessary to identify the entity in that context.
+
+It is important to note that this ITE doesn't mandate any specific generic URI
+tokens that all conforming implementations MUST support. Instead, it merely
+specifies what any generic URI must look like, and what rules the system MUST
```
lukpueh: Can you please review the document for uniform case in [key words](https://www.ietf.org/rfc/rfc2119.txt).

ITE/4/README.adoc
lukpueh: As [said above](https://github.com/in-toto/ITE/pull/6/files#r389807736), I think this is a good idea and should be mentioned in this ITE-4.

ITE/4/README.adoc
lukpueh: A change in either content or format seems to generalize to the same problem. I think you describe this well in the 3rd paragraph of the Security section below.

Maybe you can somehow add your suggestion that  *"the change itself is recorded as a step and we use its link to validate the change, or the functionary must replace existing metadata with the updated ones"*.

ITE/4/README.adoc
adityasaky: > Through separate ITEs?

We could have an ITE for each but I'm wondering if we'd have a bunch as URI tokens are added. And these aren't changes to the core specification, so is it okay to use the ITE process like that?

Doing so also allows the in-toto community to assess and contribute to how abstract resources are hashed and recorded. I think a formal and open process, even if left to the implementers, is necessary to ensure security guarantees aren't lost. I also looked into having a single, growing ITE to collect schemes and hashing information, but that doesn't seem quite possible, as ITEs aren't designed for frequent change (which makes sense). Indeed, from ITE-1:

_"Ideally, all changes to an ITE should be completed before it is "Accepted", but surprises may still occur. Changes might be minor changes, such as clarifications or typo fixes, or major changes, which would alter the intent, scope, API, or core behavior of the ITE.

All changes to an already "Accepted" ITE, must be submitted via PR as usual. In the case of major changes, the Sponsor should also communicate those changes on the mailing list and take sufficient time to ensure there is consensus on the changes before merging them. A link to any discussion should be added to the PR for the change and to the References section. If there are significant objections or questions around proposed changes, the ITE Sponsor or Reviewer may choose to return the ITE to a "Draft" status for more extensive discussion and eventual review again for acceptance."_

ITE/4/README.adoc
```diff
@@ -0,0 +1,502 @@
...
+
+[[security]]
+== Security
+
+As per the in-toto specification, the only direct interaction in-toto tools have
+with artifacts is to record their hashes using one or more cryptographic hash
+algorithms. Other artifact operations such as verifying the artifact rules
+rely on the hashes recorded in the link metadata.
+
+We found that the extending the recording of hashes to abstract resources, some
```
lukpueh:
```suggestion
We found that the recording of hashes of abstract resources, some
```

?

ITE/4/README.adoc
```diff
@@ -0,0 +1,502 @@
...
+
+[[infrastructure-requirements]]
+== Infrastructure Requirements
+
+This ITE proposes no infrastructure changes.
+
+[[testing]]
+== Testing
+
+In an in-toto system conforming to this ITE, it is important to test:
```
lukpueh: It seems hard to test this ITE, if concrete schemes and resource definitions are not in its scope.

ITE/4/README.adoc
lukpueh: You are right. ITEs don't seem to be the right place to collect these resource definitions.

ITE/4/README.adoc
adityasaky: As I understand it, it doesn't apply to every occurrence of one of the words, correct?

ITE/4/README.adoc
adityasaky: This isn't part of the RFC. Is it okay to use the same format?

ITE/4/README.adoc
lukpueh: Correct. I don't know the exact rules though. I just noticed in above two lines that there are two *MUST* and one *must* and I couldn't see the pattern. If you chose different case for a reason then it is fine. :)

ITE/4/README.adoc
lukpueh: Good catch and apologies for the nonsense suggestion. Maybe *MAY* would be the right choice of words?

ITE/4/README.adoc
adityasaky: No, I reviewed it again, I have the changes locally. During this review, I felt it didn't apply for every occurrence.

adityasaky: Hope this looks better @lukpueh :)

ITE/4/README.adoc
adityasaky: Yep. Perhaps we can plan to formalise the _process_ of adding a token and how the relevant content should be hashed in a separate ITE, and mention that here?

cc @SantiagoTorres as well

SantiagoTorres: I believe that this is probably the best path forward, and it follows the philosophy that TUF has with their wireline formats ([TAP  11](https://github.com/theupdateframework/taps/blob/master/tap11.md)).

adityasaky: Updated with:

> It is important to note that this ITE doesn’t mandate any specific generic URI tokens that all conforming implementations MUST support. Instead, it merely specifies what any generic URI MUST look like, and what rules the system MUST follow. However, it is vital to create a formal process that introduces new URI tokens and details how their abstract contents are hashed. This ensures that different in-toto implementations do not diverge significantly, and by creating an open process that involves the in-toto community, it can ensure consistency when hashing resources that multiple stakeholders use. Finally, such a process will also help ensure that the hashing of these entities is done in a correct and secure manner. However, detailing such a process is out of the scope of this ITE and must instead be handled in a future ITE.

ITE/4/README.adoc
```diff
@@ -0,0 +1,509 @@
...
+verification workflow.
+
+As always, it is also necessary to consider the actors who can make changes to
+an abstract entity. This is perhaps slightly exacerbated in the case of abstract
+resources as the content and format being hashed aren't as specific as artifacts
+in the local filesystem. In the above example, for a public repository, *any*
+GitHub user can comment on a pull request, so for an implementation that also
+considers comments when recording the hash, any user can potentially cause a
+failure of the verification workflow. This can potentially be leveraged by a
+malicious actor to target automated pipelines that rely on in-toto verification.
```
trishankatdatadog: This sentence is super vague. Can you elaborate on what attacks they can do? Is it just DoS, or can they cause malicious content to be slipped in?

ITE/4/README.adoc
trishankatdatadog: I especially want to see a working, stable implementation for GitHub that has been tested for at least a few months. Right now, the ideas sound good, but I'd like to see some empirical validation.

ITE/4/README.adoc
adityasaky: I meant DoS, I'll specify there.

ITE/4/README.adoc
adityasaky: I agree - I think proper testing needs to be addressed per resource type when we have an ITE to handle new resource types.

ITE/4/README.adoc
```diff
@@ -0,0 +1,510 @@
...
+As per the in-toto specification, the only direct interaction in-toto tools have
+with artifacts is to record their hashes using one or more cryptographic hash
+algorithms. Other artifact operations such as verifying the artifact rules
+rely on the hashes recorded in the link metadata.
+
+We found that the recording of hashes of abstract resources, some of which may
+live at remote locations, is more complicated than the recording of hashes of
+artifacts in the local filesystem. It is, therefore, important for ITE-4
+compliant in-toto tools to handle the resolving of generic URIs to these
+abstract resources securely. It is quite possible that the contents of these
```
JustinCappos: What does "securely" mean here?  What if a resource is unavailable?

ITE/4/README.adoc
```diff
@@ -0,0 +1,510 @@
...
+and these operations when performed unsafely can lead to severe vulnerabilities
+if the resource is controlled by a malicious actor.
+
+Further, it is likely that abstract resources change more frequently, both in
+content and format, and implementors must take care to identify how these
+contents are recorded, as well what specific information is recorded for a
+particular type of entity. Otherwise, the in-toto verification workflow can be
+plagued by failures due to the lack of availability of the artifact as
+previously recorded. For example, implementors who decide how to resolve GitHub
+pull request must decide what information encoded in a pull request must be
```
JustinCappos: Can you just state that you require such items to be static (non-changing) and declare modifiable items to be out of scope?

ITE/4/README.adoc
```diff
@@ -0,0 +1,510 @@
...
+in the local filesystem. In the above example, for a public repository, *any*
+GitHub user can comment on a pull request, so for an implementation that also
+considers comments when recording the hash, any user can potentially cause a
+failure of the verification workflow. This can potentially be leveraged by a
+malicious actor to target automated pipelines that rely on in-toto verification,
+using a Denial of Service attack.
+
+Finally, our analysis showed that verification of artifact rules specified in
+in-toto layouts use the hashes recorded while generating the link metadata and
+do not record afresh any of the hashes. So the security properties of artifact
```
JustinCappos: What does the word 'security' mean here?

ITE/4/README.adoc
adityasaky: > Can you just state that you require such items to be static (non-changing) and declare modifiable items to be out of scope?

I may be overthinking here, but do you mean that these resources shouldn't change _in_ the supply chain? Meaning that `MODIFY` doesn't apply to these resources, and any change would require links to be regenerated?

ITE/4/README.adoc
adityasaky: What I'm trying to convey is that we don't have anything to worry about just from verification of rules as long as all the previous points about resolving are implemented - being careful with serializing, changes in abstract resources, (now) outright missing resources, and being careful with repeatedly changing resources (such as comments on a PR). This will of course be further affected / limited depending on the other discussion about limiting scope to static resources. I may be using the wrong words here.

ITE/4/README.adoc
adityasaky: I meant that they must be careful with how they serialize. Additionally, do you think it's okay to add that implementers must handle situations where resources are unavailable by failing appropriately, as the context demands?

adityasaky:
Updated with a link to #7 in the document. I think this version can be merged with the current status?

@SantiagoTorres

ITE/4/README.adoc
```diff
@@ -0,0 +1,598 @@
...
+define hash_artifacts(string generic_uri, list hash_algorithms, kwargs) \
+    returns map artifact_hash_map
+
+define resolve_uri(string generic_uri, kwargs) returns list contained_uris
+
+define get_hashable_representation(string generic_uri, kwargs) returns bytes \
+    hashable_representation
+```
+
+`hash_artifact` closely mirrors the general workflow of hashing artifacts in the
```
trishankatdatadog:
```suggestion
`hash_artifacts` closely mirrors the general workflow of hashing artifacts in the
```

adityasaky: Fixed, thanks for the catch. :)

ITE/4/README.adoc
```diff
@@ -0,0 +1,598 @@
...
+
+==== General Structure for URI Resolvers
+
+While the implementations of resolvers will vary wildly depending on the context
+they're written for, this ITE recommends they follow a certain structure that
+enables them to be reused by different stakeholders looking to hash the same
+abstract entities.
+
+```
+define hash_artifacts(string generic_uri, list hash_algorithms, kwargs) \
```
trishankatdatadog: Maybe say `function` instead of `define`? Could be clearer.

ITE/4/README.adoc
```diff
@@ -0,0 +1,598 @@
...
+a particular entity and they must be generated keeping in mind the different
+sources of non-determinism as highlighted in the Appendix. It's also necessary
+for this method to handle missing resources safely.
+
+Being able to get some hashable representation of abstract entities is necessary
+for in-toto inspections to verify contents as the context demands it. Therefore,
+it should be possible for implementations to use the method in scenarios other
+than just the recording of hashes of an artifact.
+
+It is quite likely that a generic URI specified may point to a collection of
```
trishankatdatadog: Could you explain how these functions relate with each other? For example, who is supposed to call whom? Maybe a small toy example would help.

adityasaky: Do you think 8fc095b clarifies enough?

ITE/4/README.adoc
JustinCappos: Can you change the word in the text so that you are more specific?

ITE/4/README.adoc
JustinCappos: I mean that the thing the URI points to should be static, not that you cannot have a MODIFY step or similar.

ITE/4/README.adoc
JustinCappos: Okay, can you clarify in the text please?

adityasaky: I've updated it to read since I think the exact handling of a missing resource may be context dependent:

> Implementors must also take care to ensure they handle situations where a resource may be unavailable, and fail appropriately.

Do you think we need to be more strict and say the workflow must fail? I was also thinking of inspections that explicitly check if a resource is unavailable to pass.

ITE/4/README.adoc
adityasaky: I opened #7 and I've inserted a link in the document to further discuss this, based especially on my findings with implementing this for GitHub.

ITE/4/README.adoc
adityasaky: I rewrote this bit to read:

> Finally, our analysis showed that verification of artifact rules specified in in-toto layouts rely on the hashes recorded while generating link metadata for steps and inspections, and do not record any hashes afresh. Therefore, while the changes proposed in this ITE impact the resolving of artifacts and the recording of their hashes, they do not change how artifact rules are verified.

ITE/4/README.adoc
JustinCappos: CC: @SantiagoTorres 

I'm not sure what makes sense.  Should every in-toto implementation make its own decision?  Should the layout specify what should happen in this case?  Should the user be prompted (whatever that means)?

SantiagoTorres: The workflow I was imagining/hoping-for was to have resolvers exist like plug-ins for implementations. I think we should `MUST` fail if it needs to resolve a resource but doesn't have the code to do so makes sense to me.

I didn't think of having information in the layout to say what should happen if that's the case. I think it may be useful but I wonder if it would make the layout too verbose...

lukpueh: I support clearly defined behavior such as "MUST fail" and not making the layout more verbose.

I also think it's worth to stress once more in this section that the availability of resources is *only* relevant for resources that are hashed and recoded in the course of running inspections, right?

Also note that the reference implementation currently does not allow to parametrize which materials and products should be hashed and recorded for inspections, instead we record all files in the verification cwd.

adityasaky: > I think we should MUST fail if it needs to resolve a resource but doesn't have the code to do so makes sense to me.

I agree with this, but this section is talking about compliant tools, meaning they presumably have the means to resolve the URI but the resource is missing at that location. That, I believe, should be handled per context, per type of resource.

SantiagoTorres: I think we can rephrase and clarify. I can see how many readers could read what Justin read from what is there.

adityasaky: I think https://github.com/in-toto/ITE/pull/6/commits/de11c18cac1fc07be8f4013422694c359c369deb clarifies that.

SantiagoTorres:
At this point, I think we can merge as draft and continue the discussion in the mailing list with broader feedback from the community

PR #9, opened 2020-05-05
ITE-4: Expand motivation 3 to CNAB bundles; add ITE-4 to README's drafts section

adityasaky:
Also updates the prototype implementation section to point to basic proof-of-concept resolver for CNAB bundle.json files.

### DIFF
--- a/ITE/4/README.adoc
+++ b/ITE/4/README.adoc
@@ -441,7 +441,9 @@ A practical example of this use case is for Cloud Native Application Bundles
 contains information such as the top-level package information, images included
 in the bundle, credentials required by the application, and so on. We define a
 minimal supply chain described below for the creation of these bundle.json
-files. A link:https://github.com/adityasaky/in-toto/tree/ite-4-cnab-resolver[prototype resolver]
+files. This supply chain is used in CNAB Signy and can be
+link:https://github.com/cnabio/signy/blob/master/scripts/in-toto/minimal.py[verified using in-toto].
+A link:https://github.com/adityasaky/in-toto/tree/ite-4-cnab-resolver[prototype resolver]
 was created using the structure defined in the <<uri-resolvers, URI Resolvers>>
 section for CNAB bundle.json files.
 
@@ -618,6 +620,8 @@ the workflow laid out in Use Case 3 for CNAB bundles.
 * link:https://www.ntia.doc.gov/SoftwareTransparency[NTIA Software Component Transparency]
 * link:https://www.it-cisq.org/software-bill-of-materials/index.htm[CISQ/OMG Tool-to-tool Software Bill of Materials Exchange]
 * link:https://cnab.io/[Cloud Native Application Bundles]
+* link:https://github.com/cnabio/signy/blob/master/scripts/in-toto/minimal.py[Minimal Supply Chain for CNAB bundle.json files]
+* link:https://github.com/adityasaky/in-toto/tree/ite-4-cnab-resolver[CNAB Prototype Resolver]
 
 [[appendix]]
 == Appendix: Sources of Non-Determinism
--- a/ITE/4/README.adoc
+++ b/ITE/4/README.adoc
@@ -437,7 +437,7 @@ the information contained in a specific key of a JSON file using JSONPath. This
 allows for greater flexibility when using in-toto with different types of
 artifacts.
 
-A practical example of this use case is for Cloud Native Application Bundle
+A practical example of this use case is for Cloud Native Application Bundles
 (CNAB) metadata. A link:https://github.com/cnabio/cnab-spec/blob/master/101-bundle-json.md[CNAB bundle.json file]
 contains information such as the top-level package information, images included
 in the bundle, credentials required by the application, and so on. We define a
@@ -615,7 +615,7 @@ the workflow laid out in Use Case 3 for CNAB bundles.
 * link:https://github.com/cdfoundation/sig-security-sbom[CD Foundation Special Interest Group on Software Bill of Materials]
 * link:https://www.ntia.doc.gov/SoftwareTransparency[NTIA Software Component Transparency]
 * link:https://www.it-cisq.org/software-bill-of-materials/index.htm[CISQ/OMG Tool-to-tool Software Bill of Materials Exchange]
-* link:https://cnab.io/[Cloud Native Application Bundle]
+* link:https://cnab.io/[Cloud Native Application Bundles]
 
 [[appendix]]
 == Appendix: Sources of Non-Determinism
--- a/ITE/4/README.adoc
+++ b/ITE/4/README.adoc
@@ -151,6 +151,7 @@ will also help ensure that the hashing of these entities is done in a correct
 and secure manner. Yet, detailing such a process is out of the scope of this
 ITE.
 
+[[artifact-rules]]
 ==== Artifact Rules
 
 in-toto artifact rules are a means to specify what artifacts the project owner
@@ -194,6 +195,7 @@ their own, making them part of the supply chain, or metadata is replaced by
 authorized functionaries. The security implications of frequently changing
 generic resources are discussed in the Security section below.
 
+[[uri-resolvers]]
 ==== URI Resolvers
 
 ITE-4 compliant in-toto tools will require pluggable components for determining what resources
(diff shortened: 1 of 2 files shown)


ITE/4/README.adoc
```diff
@@ -430,35 +432,76 @@ resulting build and CI report.
...
+contains information such as the top-level package information, images included
+in the bundle, credentials required by the application, and so on. We define a
+minimal supply chain described below for the creation of these bundle.json
+files. A link:https://github.com/adityasaky/in-toto/tree/ite-4-cnab-resolver[prototype resolver]
+was created using the structure defined in the <<uri-resolvers, URI Resolvers>>
+section for CNAB bundle.json files.
+
+===== in-toto link attestation for the creation of the initial bundle.json
+
+In this step of the minimal supply chain, a developer creates an initial
```
trishankatdatadog, line 451: Maybe link to this example root layout for clarity?

https://github.com/cnabio/signy/pull/74

ITE/4/README.adoc
```diff
@@ -430,35 +432,76 @@ resulting build and CI report.
...
+        "cnab+json:/home/saky/Work/secure-systems-lab/cnab-ite-4/bundle.json$.images.my-microservice.contentDigest": <HASH>,
+        "cnab+json:/home/saky/Work/secure-systems-lab/cnab-ite-4/bundle.json$.images.my-microservice.description": <HASH>,
+        "cnab+json:/home/saky/Work/secure-systems-lab/cnab-ite-4/bundle.json$.images.my-microservice.image": <HASH>,
+        "...": "...",
+    }
+}
+```
+
+In the example link file for this step, we notice that the in-toto tool captured
+each field (or leaf node) within the bundle.json file as a separate artifact.
```
trishankatdatadog, line 475: Maybe explain why recording only leaf nodes is sufficient to recompute the digest of intermediate and root nodes?

ITE/4/README.adoc
```diff
@@ -430,35 +432,76 @@ resulting build and CI report.
 ==== Use Case 3: Granular JSON Access
 
 The enhancement proposed in this ITE can also be used to provide more granular
-access to certain resources, such as JSON files. A generic URI can be used to
-resolve to the information contained in a specific key of a JSON file. This allows
-for greater flexibility when using in-toto with different types of artifacts.
+access to resources such as JSON files. A generic URI can be used to resolve to
+the information contained in a specific key of a JSON file using JSONPath. This
```
trishankatdatadog, line 436: Could you explain how to escape JSONPath w/o conflicting with `fnmatch` metacharacters?

ITE/4/README.adoc
adityasaky: Do you think it's better to point to it after it's reviewed and merged?

trishankatdatadog: I think it's okay to point to PR, which at least makes it clear where the relevant code is...

trishankatdatadog:
I just realized that recording leafs is not enough for JSON. Consider the case where a machine rearranges elements in a list produced by developers, where the order may change semantics.

ITE/4/README.adoc
```diff
@@ -570,6 +618,7 @@ This ITE currently proposes no prototypes.
 * link:https://github.com/cdfoundation/sig-security-sbom[CD Foundation Special Interest Group on Software Bill of Materials]
 * link:https://www.ntia.doc.gov/SoftwareTransparency[NTIA Software Component Transparency]
 * link:https://www.it-cisq.org/software-bill-of-materials/index.htm[CISQ/OMG Tool-to-tool Software Bill of Materials Exchange]
+* link:https://cnab.io/[Cloud Native Application Bundles]
```
shibumi, line 621: What about adding the other references to this section, too? For example the prototype implementation and the PR that was mentioned by @trishankatdatadog . I am a big fan of a big appendix of references :nerd_face:

trishankatdatadog:
So, in our quest to try to encapsulate [a CNAB use-case](https://github.com/cnabio/signy/pulls), @adityasaky and I learned something interesting today.

We were trying to use [JsonPath](https://github.com/json-path/JsonPath) as a way to express exactly which parts of a [CNAB](https://github.com/cnabio/cnab-spec) a machine is allowed to modify. However, we quickly found out that the metacharacters of JsonPath conflict with [fnmatch](https://docs.python.org/3/library/fnmatch.html), which is what python-in-toto uses to (partly) implement ["bash-style wildcards"](https://github.com/in-toto/docs/blob/581895c91fba2273629fa24148ec2ac9e6df8c4b/in-toto-spec.md).

The question is whether bash-style wildcards are too limiting, especially in the context of ITE-4, which is intended to generalize artifacts from beyond files to arbitrary resources with URIs. There are at least two options for fixing this:

1. ITE-4 should make it clear that no implementation must use anything other than bash-style wildcards for pattern matching.
    1. Pros: security has been fully analyzed.
    1. Cons: potentially limits interesting use cases, such as the one above.
1. Open in-toto verifiers to accepting user-specified [`filter()`](https://github.com/in-toto/docs/blob/581895c91fba2273629fa24148ec2ac9e6df8c4b/in-toto-spec.md) functions which implement arbitrary pattern matching (e.g., JsonPath).
    1. Pros: opens up interesting use cases, such as the one above.
    1. Needs thinking about security implications. For example, arbitrary regexes can open up [algorithmic complexity attacks](https://dwheeler.com/secure-programs/Secure-Programs-HOWTO/complexity-attacks.html). Furthermore, the security in-toto verification now depends on user input.

So far, this is an interesting question that we believe deserves at least some discussion. Let us know what you think here.

Cc @radu-matei @jlegrone @squillace

adityasaky:
Thanks for summarizing, @trishankatdatadog!

I'm currently leaning towards option 1, at least with the scope we've considered right now. I believe that opening up `filter()` requires far more discussion and thought, probably enough for it to be an ITE of its own? Just a thought. I also believe that swapping `filter()` is not necessarily linked to situations involving abstract resources, so it shouldn't be bundled with ITE-4. In fact, I wonder if we'd end up evaluating some popular options individually and talking about how they should be used / what considerations implementations must make etc. Such an ITE would also probably greatly benefit from the lessons learned as we continue working on ITE-4 use cases + their resolvers...

ITE/4/README.adoc
adityasaky: I've unfortunately ignored this PR a bit, sorry for the delay. I'll add links here, once we're sure about what goes into this PR. Thanks! :slightly_smiling_face:

adityasaky:
For now, I've updated the [CNAB resolver PoC](https://github.com/adityasaky/in-toto/blob/ite-4-cnab-resolver/in_toto/resolvers/cnab_json.py) to remain friendly with bash-style patterns, meaning it'll work okay with `fnmatch`, and verification passes (with `DISALLOW *` this time :wink:).

ITE/4/README.adoc
adityasaky: Resolving in favour of discussion below.

ITE/4/README.adoc
```diff
@@ -430,35 +432,80 @@ resulting build and CI report.
...
+{
+    "_type": "link",
+    "byproducts": {},
+    "command": [],
+    "environment": {},
+    "materials": {},
+    "name": "developer",
+    "products": {
+        "...": "...",
+        "cnab+json:/home/saky/Work/secure-systems-lab/cnab-ite-4/bundle.json$/images/my-microservice/contentDigest": <HASH>,
```
trishankatdatadog, line 467: I disagree that the CNAB use case can be covered by this TAP. A URI is supposed to address a document, and not necessarily parts of it, unless we refer to anchor parts (which has a standard notation). This looks like "abusing" the URI scheme to address elements inside a JSON document, but perhaps I am interpreting the meaning of the URI scheme too strictly.

adityasaky: Very technically, RFC 3986 does not apply any such limits.

```
Resource

      This specification does not limit the scope of what might be a
      resource; rather, the term "resource" is used in a general sense
      for whatever might be identified by a URI.  Familiar examples
      include an electronic document, an image, a source of information
      with a consistent purpose (e.g., "today's weather report for Los
      Angeles"), a service (e.g., an HTTP-to-SMS gateway), and a
      collection of other resources.  A resource is not necessarily
      accessible via the Internet; e.g., human beings, corporations, and
      bound books in a library can also be resources.  Likewise,
      abstract concepts can be resources, such as the operators and
      operands of a mathematical equation, the types of a relationship
      (e.g., "parent" or "employee"), or numeric values (e.g., zero,
      one, and infinity).
```

I believe that the CNAB use case must be covered by ITE-4 and such a use case was certainly intended in spirit when it was written. It may, however, be worth looking again at the proposed URI format for the CNAB use case again, as well as expanding on the general format (which just talks about `scheme` and `hier-part`) as a whole.

adityasaky: To expand on the intended in spirit bit, I think some of the same questions can be raised for artifacts in SPDX documents as well, which was also considered when writing this ITE.

SantiagoTorres:
LGTM. Thanks!

PR #13, opened 2020-09-30
ITE-5: Replace signature wrapper with SSL signing spec

MarkLodato:
This is an initial draft of ITE-5 that defines a new signature scheme. The reasoning is explained in the body. I have already discussed with @SantiagoTorres, who will sponsor this ITE.

Note: This is in Markdown because that's the format that I know and it's much simpler to edit in. We'll convert to Asciidoc before submitting. I'm sending this out now to get initial feedback on the idea.

Several sections are still empty. They need to be fleshed out (or removed) before the ITE is accepted.


ITE/5/README.md
```diff
@@ -0,0 +1,343 @@
...
+
+The switch from Hex to Base64 for `sig` is to save space and to be consistent
+with `payload`.
+
+## Steps
+
+To sign:
+
+-   Serialize BODY according to PAYLOAD_TYPE. Call the result SERIALIZED_BODY.
+-   Sign PAE([PAYLOAD_TYPE, PAYLOAD]), base64-encode the result, and store it in
```
wietse-gmail, line 87: PAYLOAD is not defined above. Did you mean: SERIALIZED_BODY? Ditto under "to verify" below.

I think that using PAYLOAD here would be confusing, because there already is a JSON field called "payload" and that field contains a transformation of SERIALIZED_BODY.

MarkLodato: Oops, thanks. Fixed.

adityasaky:
Hi Mark, thank you for putting this up as an ITE! It makes it easier for the community to discuss and evaluate proposals.

ITE-5 came up today in an in-toto + TUF discussion on our signature wrapper and our current plan to [spin it off into a standalone spec](https://github.com/secure-systems-lab/signing-spec/issues/1). While this doesn't yet mean this ITE is approved, it received positive feedback. The primary concern is with supporting two formats - we have some legacy clients that will need to continue using the old format, and that'll have to be phased out slowly. We aren't certain we have the bandwidth to maintain implementations for two formats. But apart from that, from the perspective of adopting these into the specification, there are positive signs!

MarkLodato:
Great, thank you for the update! If there is anything I can do to help, please let me know.

(Note: I'll be on leave for the next several weeks but my colleagues will be covering for me.)

trishankatdatadog:
Now that I think about it: it's good that payload type is part of each signature, but what happens when sig A claims that the serialized body is type X, but sig B claims that it is type Y? Should payload type be a fixed part of [key distribution](https://github.com/in-toto/docs/blob/c2251b3db89d9c06e07918022b58b70079c77b10/in-toto-spec.md#42-file-formats-general-principles)? Should all keys/sigs be required to have the same payload type for the serialized body?

MarkLodato:
Hi Trishank, there is only a single `payloadType` at the top level so all signatures must use the same type. Is there something in the doc that would imply otherwise? I'm happy to make edits to clarify.

Personally, I'd recommend removing multiple signature support altogether. It is a rarely used feature that adds increased complexity and attack surface. Example: go-jose had a [vulnerability](https://rwc.iacr.org/2017/Slides/nguyen.quan.pdf#page=12) where it would return "verified" if *any* of the signatures passed. If multiple signatures are needed, each signer can produce a separate link file. You can always verify that the payloads are identical, if needed.

I didn't go so far as to suggest that in this doc, but perhaps I should?

trishankatdatadog:
> Hi Trishank, there is only a single `payloadType` at the top level so all signatures must use the same type. Is there something in the doc that would imply otherwise? I'm happy to make edits to clarify.
> 
> Personally, I'd recommend removing multiple signature support altogether. It is a rarely used feature that adds increased complexity and attack surface. Example: go-jose had a [vulnerability](https://rwc.iacr.org/2017/Slides/nguyen.quan.pdf#page=12) where it would return "verified" if _any_ of the signatures passed. If multiple signatures are needed, each signer can produce a separate link file. You can always verify that the payloads are identical, if needed.
> 
> I didn't go so far as to suggest that in this doc, but perhaps I should?

in-toto produces signatures per functionary separately using different link file per signature. However, TUF shares the same signature format as in-toto, and will be affected by this change. In TUF, we use threshold signatures to require that m out of n roles have signed off on the same metadata payload. TUF does not intrinsically suffer from the same vulnerability as go-jose.

What do others think? @JustinCappos @SantiagoTorres

davidstrauss:
We would value the ability to validate signatures in the client without a cJSON implementation (or any JSON parsing at all, ideally).

Perhaps off-topic for this issue but related to signature modularity, it would also be nice to support FIDO2-style signatures, which could offer a uniquely easy and inexpensive way to enforce offline, hardware-isolated signing.

SantiagoTorres:
Hi everyone. I'm shaking the dust off this ITE. I rephrased things around to check the specifics on signing-spec, rather than here. I also tried to elaborate in our in-toto-specific motivations. Does this look like what you had in mind?

@adityasaky @MarkLodato @trishankatdatadog @JustinCappos

ITE/5/README.adoc
```diff
@@ -0,0 +1,182 @@
...
+1. avoid mandating an ad-hoc signing specification for in-toto
+2. Allow for implementers to propose different signing specifications.
+
+In adition, this ITE makes a recommendation using the
+link:http://github.com/secure-systems-lab/signing-spec[SSL Signing Spec]. Which
+has the following benefits over the current state:
+
+1. Avoids canonicalization for security reasons (i.e., to not parse untrusted input) 
+2. Reduce the possibility of misinterpretation of the payload. The serialized payload is encoded as a string and verified by the recipient before de-serializing.
+3. This specification also provides a backwards compatible serialization format.
```
trishankatdatadog, line 52:
```suggestion
3. This specification also provides a backwards-compatible serialization format.
```

ITE/5/README.adoc
```diff
@@ -0,0 +1,182 @@
...
+JSON-specific.) We believe the SSL signing spec strikes the right balance of
+simplicity, usefulness, and security. 
+
+Further, the SSL signing spec is a "natural evolution" of the current signature
+wrapper, which was both defined in the TUF and in-toto specifications. As such,
+it allows for a transparent upgrade via their cryptographic provider,
+link:https://github.com/secure-systems-lab/securesystemslib[securesystemslib].
+
+Finally, since this is an improved version of the current TUF and in-toto
+wrappers, the design allows for a backwards compatible JSON representation,
```
trishankatdatadog, line 97:
```suggestion
wrappers, the design allows for a backwards-compatible JSON representation,
```

ITE/5/README.adoc
```diff
@@ -0,0 +1,182 @@
...
+[[Security]]
+== Security
+
+At first sight this proposal is central to security, yet the actual
+contribution is to allow for a signature provider to be disaggregated from the
+specification. As such, no supply-chain security properties are removed from
+the system through this ITE.
+
+The specifics on the signature wrapper provide, at worst, the same security
+properties through its backwards compatible json representation. At best, the
```
trishankatdatadog, line 113:
```suggestion
properties through its backwards-compatible json representation. At best, the
```

ITE/5/README.adoc
```diff
@@ -0,0 +1,182 @@
...
+
+|===
+
+[[abstract]]
+== Abstract
+
+This link:https://github.com/in-toto/ITE[in-toto enhancement (ITE)] proposes
+removing the in-toto signature scheme requirements, and allow users to adopt the signature wrapper that better fits their ecosystem. This has the following general benefits:
+
+1. avoid mandating an ad-hoc signing specification for in-toto
```
adityasaky, line 43:
```suggestion
1. avoid mandating an ad-hoc signing specification for in-toto; and
```

ITE/5/README.adoc
```diff
@@ -0,0 +1,182 @@
...
+2. Allow for implementers to propose different signing specifications.
```
adityasaky, line 44:
```suggestion
2. allow for implementers to propose different signing specifications.
```

ITE/5/README.adoc
```diff
@@ -0,0 +1,182 @@
...
+
+In adition, this ITE makes a recommendation using the
+link:http://github.com/secure-systems-lab/signing-spec[SSL Signing Spec]. Which
```
adityasaky, line 47:
```suggestion
link:http://github.com/secure-systems-lab/signing-spec[SSL Signing Spec]. This
```

ITE/5/README.adoc
```diff
@@ -0,0 +1,182 @@
...
+[[abstract]]
+== Abstract
+
+This link:https://github.com/in-toto/ITE[in-toto enhancement (ITE)] proposes
+removing the in-toto signature scheme requirements, and allow users to adopt the signature wrapper that better fits their ecosystem. This has the following general benefits:
+
+1. avoid mandating an ad-hoc signing specification for in-toto
+2. Allow for implementers to propose different signing specifications.
+
+In adition, this ITE makes a recommendation using the
```
adityasaky, line 46:
```suggestion
In adition, this ITE makes a recommendation to use the
```

ITE/5/README.adoc
```diff
@@ -0,0 +1,182 @@
...
+link:http://github.com/secure-systems-lab/signing-spec[SSL Signing Spec]. Which
+has the following benefits over the current state:
+
+1. Avoids canonicalization for security reasons (i.e., to not parse untrusted input) 
+2. Reduce the possibility of misinterpretation of the payload. The serialized payload is encoded as a string and verified by the recipient before de-serializing.
```
adityasaky, line 51:
```suggestion
2. Reduces the possibility of misinterpretation of the payload. The serialized payload is encoded as a string and verified by the recipient before de-serializing.
```

ITE/5/README.adoc
```diff
@@ -0,0 +1,182 @@
...
+wrappers (i.e., one for TUF and one for in-toto) is rather difficult. As time
+passes, and due to the interactions in each community, the signature wrappers
+evolve to better fit their user-bases. Adopting a common source of truth (i.e.,
+a separate signature specification) should help increase cohesion between
+projects while maintaining the original goal of code-reuse and transparent
+integration.
+
+In addition, keeping the signature wrapper specification *outside* of the
+current in-toto specification will also simplify the specification, which can
+now only focus in describing in-toto specifics, rather than cryptographic
```
adityasaky, line 79:
```suggestion
now only focus on describing in-toto specifics, rather than cryptographic
```

adityasaky (approved):
Minor nits, but this looks good to me! Can we get rid of the hypothetical attack?

ITE/5/README.adoc
adityasaky: Or "this ITE recommends using" I guess.

SantiagoTorres:
thank you @adityasaky !

ITE/5/README.adoc
```diff
@@ -0,0 +1,182 @@
...
+
+[[prototype-implementation]]
+== Prototype Implementation
+
+See link:https://github.com/secure-systems-lab/signing-spec/blob/master/reference_implementation.ipynb[reference_implementation.ipynb] for an implementation.
+
+[[Testing]]
+== Testing
+
+=== Test Vector
```
TomHennen, line 132: Would it make more sense to only have this test vector in the SSL signing spec and then _not_ reference it here?

(I'm think this isn't really in-toto specific at this point so it can be delegated to the signing spec.)

SantiagoTorres: Hmm, fair point. Let me move this around.

ITE/5/README.adoc
```diff
@@ -0,0 +1,182 @@
...
+[[motivation]]
+== Motivation
+
+The early motivations for in-toto to use the current wrapper were to allow for
+our sister project, TUF, to transparently use the same crypto provider. The
+reasoning back then is that we would maximize code reuse, and allow for users
+of both solutions to easily integrate them.
+
+However, as time has shown, keeping these two separately-specified signature
+wrappers (i.e., one for TUF and one for in-toto) is rather difficult. As time
```
joshuagl, line 70:
```suggestion
wrappers (i.e., one for TUF and one for in-toto) synchronized is rather difficult. As time
```

ITE/5/README.adoc
```diff
@@ -0,0 +1,182 @@
...
+== Reasoning
+
+Our goal was to adopt a signature wrapper that is as simple and foolproof as
+possible. Alternatives such as [JWS] are extremely complex and error-prone,
+while others such as [PASETO] are overly specific. (Both are also
+JSON-specific.) We believe the SSL signing spec strikes the right balance of
+simplicity, usefulness, and security. 
+
+Further, the SSL signing spec is a "natural evolution" of the current signature
+wrapper, which was both defined in the TUF and in-toto specifications. As such,
```
joshuagl, line 92:
```suggestion
wrapper, which was defined in both the TUF and in-toto specifications. As such,
```

ITE/5/README.adoc
```diff
@@ -0,0 +1,182 @@
...
+[[Security]]
+== Security
+
+At first sight this proposal is central to security, yet the actual
+contribution is to allow for a signature provider to be disaggregated from the
+specification. As such, no supply-chain security properties are removed from
+the system through this ITE.
+
+The specifics on the signature wrapper provide, at worst, the same security
+properties through its backwards-compatible json representation. At best, the
```
joshuagl, line 113:
```suggestion
properties through its backwards-compatible JSON representation. At best, the
```

ITE/5/README.adoc
```diff
@@ -0,0 +1,182 @@
...
+changes improve the security stance of implementations since it is not parsing
+untrusted input anymore.
```
joshuagl, line 115:
```suggestion
changes improve the security stance of implementations since they are no longer parsing
untrusted input.
```

joshuagl (commented):
Very minor edits suggested for clarity. Apologies if this is too close to bike shedding.

MarkLodato:
It would be helpful to merge the PR so that we can link to ITE-5 without referencing this PR. It seems like there hasn't been any updates or comments in many months. Is anything more needed? Note that it's still Draft Status.

SantiagoTorres:
+1, I think there are two small comments and we could merge...

ITE/5/README.md
```diff
@@ -0,0 +1,343 @@
...
+    with a "_type" field? Then if you want a non-JSON payload, you could simply
+    have a field that contains the real payload, e.g. `{"_type":"my-thing",
+    "value":"base64…"}`.
+
+    1.  It encourages users to add a "_type" field to their payload, which in
+        turn:
+        -   (a) Ties the payload type to the authentication type. Ideally the
+            two would be independent.
+        -   (b) May conflict with other uses of that same field.
+        -   (c) May require hte user to specify type multiple times with
```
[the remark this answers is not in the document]
MarkLodato: (This is now obsolete)

ITE/5/README.adoc
```diff
@@ -0,0 +1,182 @@
...
+
+1. Avoids canonicalization for security reasons (i.e., to not parse untrusted input) 
+2. Reduces the possibility of misinterpretation of the payload. The serialized payload is encoded as a string and verified by the recipient before de-serializing.
+3. This specification also provides a backwards-compatible serialization format.
+
+[[specification]]
+== Specification
+
+The specification adopted will be the SSL signing spec, as linked above. As
+such, we defer to that document to describe the specifics of signature
```
MarkLodato, line 59: Does this ITE need to explain how it will work within in-toto? In particular, the intro above says that in-toto won't require a specific envelope but will merely recommend this as one of many that are supported. Does that need to be defined here, or will that be an implementation detail? (I'm not familiar with the ITE process.)

ITE/5/README.adoc
```diff
@@ -0,0 +1,153 @@
...
+| 2020-09-28
+
+|===
+
+[[abstract]]
+== Abstract
+
+This link:https://github.com/in-toto/ITE[in-toto enhancement (ITE)] proposes
+removing the in-toto signature scheme requirements, and allow users to adopt the
+signature envelope that better fits their ecosystem. This has the following
```
MarkLodato, line 42: This ITE doesn't seem to actually talk about how that will work - that is, how it will support arbitrary envelope types. Should we instead focus this ITE purely on supporting SSLSS (we need a better name!) and talk about how that will be implemented in in-toto?

ITE/5/README.adoc
```diff
@@ -0,0 +1,153 @@
...
+[[specification]]
+== Specification
+
+The specification adopted will be the SSL signing spec, as linked above. As
+such, we defer to that document to describe the specifics of signature
+generation and verification.
+
+The envelope's `payloadType` is either:
+
+* `https://in-toto.io/Link/v1+json` (for links)
```
MarkLodato, line 64: @SantiagoTorres What do you think about these URLs?

MarkLodato (commented):
Oops! I didn't realize there were unresolved comments. I went through and addressed them as well as made several more fixes. I think it's in OK state now, modulo the comments below.

ITE/5/README.adoc
MarkLodato: (Merging threads): Should we instead focus this ITE purely on supporting SSLSS (we need a better name!) and talk about how that will be implemented in in-toto?

ITE/5/README.adoc
MarkLodato: (This is a dupe of the other thread. Let's continue discussion there.)

ITE/5/README.adoc
```diff
@@ -0,0 +1,153 @@
...
+
+In addition, this ITE makes a recommendation to use the
+link:http://github.com/secure-systems-lab/signing-spec[SSL Signing Spec]. This
+has the following benefits over the current state:
+
+1. Avoids canonicalization for security reasons (i.e., to not parse untrusted input) 
+2. Reduces the possibility of misinterpretation of the payload. The serialized payload is encoded as a string and verified by the recipient before de-serializing.
+
+[[specification]]
+== Specification
```
MarkLodato, line 56: Reference specific signing spec version number.

ITE/5/README.adoc
MarkLodato: Santiago mentioned possible adding "/schemas" to the start.

Another question: should we have just one URI for both types, since the `_type` field within the payload indicates the type?

ITE/5/README.adoc
MarkLodato: After discussing over video, I'll add a "processing model" that discusses how to process the Envelope layer in pseudocode. That way implementations know what they should do.

ITE/5/hypothetical_signature_attack.ipynb
```diff
@@ -0,0 +1 @@
...ThmYzFjMTQ5YWZiZjRjODk5NmZiOTI0MjdhZTQxZTQ2NDliOTM0Y2E0OTU5OTFiNzg1MmI4NTVqYnlwcm9kdWN0c6BlX3R5cGVkbGluaw==',\n"," 'payloadType': 'Protobuf',\n"," 'signatures': [{'sig': 'LasJ/aXjyKdiVSNrA5uXTiH20D6Am7xa67nBEI9K6ZQYLBitn1NVhMpMEGY6QW7Qnyi6N/LKgZhLsAA5Mvur3Q=='}]}"]},"metadata":{"tags":[]},"execution_count":24}]},{"cell_type":"markdown","metadata":{"id":"6ClrxXmyqv7l"},"source":["Finally, send it to the victim and profit!"]},{"cell_type":"code","metadata":{"id":"mvxELrI0qzHI","executionInfo":{"status":"ok","timestamp":1601045411469,"user_tz":240,"elapsed":11178,"user":{"displayName":"Mark Lodato","photoUrl":"https://lh3.googleusercontent.com/a-/AOh14Gje3VCiWTKvjKn9WNmOK7z5cDhpwtiSncwf3Flh=s64","userId":"14555828759934874531"}},"outputId":"6622bca7-ee77-463b-d6a7-ce7173837a1c","colab":{"base_uri":"https://localhost:8080/"}},"source":["VerifyAndPrint(link_modified)"],"execution_count":null,"outputs":[{"output_type":"stream","text":["Good signature\n","effective_command: \"echo \\\"hello world\\\"\"\n","products {\n","  resource_uri: \"stdout\"\n","  hashes {\n","    sha256: \"badbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadb\"\n","  }\n","}\n","\n"],"name":"stdout"}]}]}
```
adityasaky, line 1: I believe this attack is only referred to in the motivation of signing-spec, so can we remove this file here? Or will the ITE use it in-text?

ITE/5/README.adoc
MarkLodato: Done. Removed the wording about supporting arbitrary envelopes and added pseudocode.

ITE/5/README.adoc
MarkLodato: Done.

ITE/5/hypothetical_signature_attack.ipynb
MarkLodato: Oops! Good catch. Removed both notebooks.

ITE/5/README.adoc
MarkLodato: Done. There is now a single URI and it includes as "schemas/" path prefix. Please take a look.

MarkLodato:
OK, I think all the comments should be resolved now. Please take a look at the latest wording, particularly the psuedocode. Thanks!

README.md
```diff
@@ -9,6 +9,7 @@
 
 * [ITE-2: A general overview of combining TUF and in-toto to build compromise-resilient CI/CD](ITE/2/README.adoc)
 * [ITE-3: Real-world example of combining TUF and in-toto for packaging Datadog Agent integrations](ITE/3/README.adoc)
+* [ITE-5: Replace signature envelope with SSL signing spe](ITE/5/README.adoc)
```
davidstrauss, line 12: spec*

MarkLodato: Oops! Fixed. Thanks for the catch.

SantiagoTorres:
Merging now :+1: I'll announce this on the ML soon™

Issue #7, opened 2020-03-24
ITE-4: Limiting scope of generic URIs to static resources


adityasaky:
In a review of PR #6 for ITE-4, @JustinCappos raised some important points about situations where the URIs (that the ITE describes) point to dynamic resources. The Security section of the draft highlights that care must be taken when recording dynamic contents. The suggestion is that this ITE should instead limit the use of generic URIs to static resources. I'm now working on identifying the impact this will have on some of the use cases that motivated this ITE.

Original discussion thread (for context): https://github.com/in-toto/ITE/pull/6#discussion_r395296284

cc @SantiagoTorres @trishankatdatadog @lukpueh

adityasaky:
We've specifically addressed the pitfalls that can occur with non static resources and included warnings. The accepted version of the ITE can now be found at https://github.com/in-toto/ITE/blob/master/ITE/4/README.adoc.

PR #14, opened 2020-10-05
ITE-4: Review for acceptance process

adityasaky:
WIP: addressing #7

### DIFF
--- a/ITE/4/README.adoc
+++ b/ITE/4/README.adoc
@@ -23,7 +23,7 @@ endif::[]
 | link:https://github.com/santiagotorres[Santiago Torres-Arias]
 
 | Status
-| Draft :speech_balloon:
+| Accepted
 
 | Type
 | Standards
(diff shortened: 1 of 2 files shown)


ITE/4/README.adoc
```diff
@@ -546,19 +548,20 @@ with artifacts is to record their hashes using one or more cryptographic hash
...
-abstract resources in a secure manner. It is quite possible that the contents of these
-resources need to be serialized in a manner that allows for their hashing, and these
-operations, when performed unsafely, can lead to severe vulnerabilities if the
-resource is controlled by a malicious actor. Implementers must also take
-care to ensure ITE-4 compliant systems are capable of handling situations where
-an abstract resource is unavailable, and fail appropriately.
-
-NOTE: There is an ongoing discussion about limiting the scope of this ITE to
-static or unchanging resources. Thread: https://github.com/in-toto/ITE/issues/7
+abstract resources in a secure manner. It is quite possible that the contents of
```
trishankatdatadog, line 555: Good advice, but how? Where should the non-expert reader turn to for best secure practices?

ITE/4/README.adoc
```diff
@@ -234,27 +237,27 @@ properties. For example, an implementation that is capable of hashing GitHub
...
+a particular entity, and to acknowledge the different sources of non-determinism
+highlighted in the Appendix. It is also necessary for this method to handle any
+missing resources safely. That is, if a resource does not exist at the location
+pointed to by a URI, this method must fail appropriately.
+
+Being able to get some hashable representation of abstract entities is also
+needed for in-toto inspections to verify contents. There is no specification
+mandating any specific inspections. Yet, these abstract resources still need to
+be resolvable so that they can be inspected if the supply chain owner so decides
+to. Therefore, it should be possible for implementations to use
```
trishankatdatadog, line 249: How does this sentence follow from the previous one? Not clear to me, at least.

ITE/4/README.adoc
```diff
@@ -132,31 +133,32 @@ follows:
...
+It is important to note that this ITE specifies what any generic URI MUST look
+like, and what rules the system MUST follow, but does not mandate a particular
+URI token that all conforming implementations MUST support. However, it is vital
+to create a formal process that introduces new URI tokens and details how their
+abstract contents are hashed. This ensures that different in-toto
+implementations do not diverge significantly. Creating an open process that
+involves the in-toto community can ensure consistency when hashing resources
+used by multiple stakeholders. Finally, such a process will also help ensure
+that the hashing of these entities is done in a correct and secure manner. Yet,
+detailing such a process is out of the scope of this ITE.
```
trishankatdatadog, line 153: Echoing my later comment here, I don't feel that this is good enough. We should at least briefly discuss what "correct" and "secure" mean, and point to an example.

ITE/4/README.adoc
adityasaky: It's very contextual as it depends on each independent type of resource, and the corresponding best practices. I've updated to indicate that, with some general examples.

ITE/4/README.adoc
adityasaky: I think I've clarified this one now. Basically, we're stating that `get_hashable_representation` should be callable on demand, for things like inspections.

ITE/4/README.adoc
adityasaky: As before, the process would establish a way to "register" a type of abstract resource. Using that process for each type of resource, we'd also establish the best practices to deal with their entities. I've mentioned it's contextual, and I've updated the security section with some general examples, where your other related comment was.

trishankatdatadog (approved):

adityasaky:
Thanks, @trishankatdatadog!

SantiagoTorres (approved):
Great! Thank you @trishankatdatadog , for your dilligent review, and @adityasaky for putting this together!

PR #15, opened 2020-10-30
ITE-6: Generalized link format

TomHennen:
This is an initial draft of ITE-6 that defines a generalized link format.  See the document for reasoning.

This has been discussed with @SantiagoTorres who will be sponsoring this ITE.

This was authored primarily by @MarkLodato (who is on leave).  Most credit goes to him.

I've converted it to Markdown, any formatting issues are likely my fault.

Several sections are still empty. They need to be fleshed out (or removed) before the ITE is accepted.


ITE/6/README.md
```diff
@@ -0,0 +1,929 @@
...
+    *   What artifacts this attestation is about.
+*   **Details**: Additional data, specified with a type-dependent schema.
+
+The binding and details together are called the **claim**.
+
+![drawing](attestation_parts.png)
+
+## Authentication and serialization
+
+[ITE-5](https://github.com/MarkLodato/ITE/blob/ite-5/ITE/5/README.md) defines a
```
TomHennen, line 70: NOTE: If accepted this will be the signing spec (https://github.com/secure-systems-lab/signing-spec/pull/2) instead of ITE-5.

ITE/6/README.md
```diff
@@ -0,0 +1,929 @@
...
+case, the payload is always JSON with a `_type` of
+“https://in-toto.io/Attestation/v1”.
+
+## Claim schema (binding layer)
+
+A claim is a JSON or CBOR object corresponding to the binding layer. It contains
+the following fields. See subsequent sections for
+[Type definitions](#type-definitions) and [Reasoning](#reasoning).
+
+`attestation_type` _(URI, required)_
```
TomHennen, line 91: It might be useful to add a story about how people add new attestation types.  E.g. how do they define the type and add the policy evaluation code.

Thoughts?

ITE/6/README.md
```diff
@@ -0,0 +1,929 @@
...
+
+> The meaning of the key is dependent on `attestation_type`. Each set value is
+> an _Object_ representing a single software artifact. The schema is dependent
+> on `attestation_type` and can contain arbitrary metadata, but it must always
+> have at least the following field:
+>
+> `artifact` _(ArtifactReference, required)_
+>
+> > Identifies the related software artifact. The following standard relations
+> > are recommended[^1]: * `top_level_source`: The primary input used to
```
TomHennen, line 155: From @MarkLodato "I'm [...] wondering whether we should have different "classes" of attestations, notably Provenance, CodeReview, TestResult (including vuln scan), and PolicyDecision, each with their own set of relations."

ITE/6/README.md
```diff
@@ -0,0 +1,892 @@
...
+## Claim schema (binding layer)
+
+A claim is a JSON or CBOR object corresponding to the binding layer. It contains
+the following fields. See subsequent sections for
+[Type definitions](#type-definitions) and [Reasoning](#reasoning).
+
+`attestation_type` _(URI, required)_
+
+> Indicates the meaning of this attestation and how to interpret `details` and
+> `relations`. Example:
```
lukpueh, line 93: There is quite a bit of forward declaration for fields and types in this section. It would be nice, to try to re-structure so that terms are defined before they are used as part of other definitions.

MarkLodato:
FYI, I just pushed a pretty significant rewrite, based on feedback from Santiago and others. It is now proposes much more modest changes from current in-toto, keeping `materials` and `products` (now called `subject`, with slightly different semantics) and making all the other fields type-specific.

The design still needs more work, but hopefully it is improving! :-D

ITE/6/spec/sbom.md
```diff
@@ -0,0 +1,48 @@
...
+
+This allows to represent an "exportable" or "published" software artifact. It
+can also be used as an entry point for other types of in-toto attestation when
+performing policy decisions
+
+## Schema
+
+```jsonc
+{
+  "attestation_type": "https://in-toto.io/Provenance/v1",
```
TomHennen, line 15: Should 'attestation_type' say something about SBOM instead of 'Provenance' ?

ITE/6/README.md
```diff
@@ -0,0 +1,574 @@
...
+
+*   Make it possible to write policies (layouts) that take advantage of
+    structured information.
+
+# Specification
+
+## Introduction
+
+An **attestation** is the generalization of an in-toto link. It is a statement
+about an artifact, signed by an attester. Each attestation has a type indicating
```
tiziano88, line 38: Does it need to be about one specific artifact? What about build steps that produce multiple ones? Does each of them end up having a separate attestation?

ITE/6/README.md
```diff
@@ -0,0 +1,574 @@
...
+`"https://in-toto.io/Attestation/v1-json"`.
+
+## Schema
+
+*   [Attestation v1](spec/attestation.md) (base class)
+*   [Provenance v1](spec/provenance.md)
+*   [Link v1](spec/link.md) (matches [in-toto 0.9])
+
+We expect to a few more standard attestation types over time, and customers may
+define their own custom attestation types if desired.
```
tiziano88, line 80: Is there a mechanism for registering types? Given they are represented by URIs / URLs, is there a meaning behind what these resolve to (if at all)?

ITE/6/README.md
```diff
@@ -0,0 +1,574 @@
...
+
+See [examples](#examples) to get a gist of the schema.
+
+## Design notes
+
+Attestations SHOULD be designed to encourage policies to be "monotonic," meaning
```
tiziano88, line 86: +1 for monotonicity

ITE/6/README.md
```diff
@@ -0,0 +1,574 @@
...
+    sources vs build tools.
+    *   Should be type-dependent, rather than mandating "materials" and
+        "products."
+
+# Reasoning
+
+## Reason for `subject` and `materials` to be standard Attestation fields
+
+There are two main reasons for standardizing `subject` and `materials` within
+Attestation schema.
```
tiziano88, line 516: Did I miss the fact that these are in fact standardized? Should this be mentioned earlier on?

ITE/6/README.md
```diff
@@ -0,0 +1,574 @@
...
+
+First, doing so allows policy engines to make decisions without requiring
+`attestation_type`-specific logic or configuration. Binary Authorization
+policies today are purely about "does an attestation exist that is signed by X
+with subject Y", and similarly in-toto layouts are about "does an attestation
+exist that is signed by X with materials/products Z?"[1] These relatively simple
+policies are quite powerful. With this proposal, such policies become more
+expressive without any additional configuration: "does an attestation exist that
+is signed by X having type T, with subject Y and/or materials Z?"
```
tiziano88, line 525: Another way of looking at this: an attestation is a partial definition of a directed graph, in which edges connect sources materials to artifacts (to use your terminology), and the edges themselves are labelled with a pointer to the attestation that introduced them. Then these attestation compose naturally, and it is also possible to traverse this graph forwards or backwards, also taking into account the trust model in the various principals that sign these attestations.

This also produces a nice way of formulating what a "root of trust" is: from a given node in the graph, follow links backwards for attestations that are trusted (e.g. "GitHub actions"), until you reach an attestation that you cannot decompose any further (because you don't trust the upstream attestations).

If we go with this semantics, effectively an attestation with a set of N materials and a set of M subjects corresponds to (N * M) edges that contribute to the graph. This assumes all the materials contribute to all the subjects, which I think is a safe assumption; if this is not desired, then perhaps the attestation should be denormalized into individual edges, in order to offer more precision, at the expense of higher verbosity and lower entropy for the more common cases.

tiziano88: BTW I think this also would connect nicely with @dlorenc 's ideas on how to incorporate more expressive entries in sigstore / rekor that contain links to / from other log entries.

dlorenc: > BTW I think this also would connect nicely with @dlorenc 's ideas on how to incorporate more expressive entries in sigstore / rekor that contain links to / from other log entries.

That's no coincidence! @SantiagoTorres was behind much of that thinking :)

ITE/6/README.md
```diff
@@ -0,0 +1,574 @@
...
+    "pkg:deb/debian/stunnel4@5.50-3?arch=amd64":               { "sha256": "e1731ae217fcbc64d4c00d707dcead45c828c5f762bcf8cc56d87de511e096fa" },
+    "pkg:deb/debian/python-impacket@0.9.15-5?arch=all":        { "sha256": "71fa2e67376c8bc03429e154628ddd7b196ccf9e79dec7319f9c3a312fd76469" },
+    "pkg:deb/debian/libzstd-dev@1.3.8+dfsg-3?arch=amd64":      { "sha256": "91442b0ae04afc25ab96426761bbdf04b0e3eb286fdfbddb1e704444cb12a625" },
+    "pkg:deb/debian/libbrotli-dev@1.0.7-2+deb10u1?arch=amd64": { "sha256": "05b6e467173c451b6211945de47ac0eda2a3dccb3cc7203e800c633f74de8b4f" }
+  },
+  // Provenance-specific fields:
+  "builder": { "id": "https://github.com/Attestations/GitHubHostedActions@v1" },
+  "recipe": {
+    "type": "https://github.com/Attestations/GitHubActionsWorkflow@v1",
+    "material": "git+https://github.com/curl/curl@curl-7_72_0",
```
tiziano88, line 131: Is there an implicit or explicit assumption somewhere that these materials that are part of the custom fields also need to be listed as part of the top-level materials? Either way it seems a bit repetitive and error prone, perhaps this can be simplified / normalized with a more structured schema.

ITE/6/README.md
TomHennen: We had thought there would be one attestation per artifact, but our work with Busytown made us realize we need to support multiple artifacts per-attestation.  This does make writing good policy somewhat harder.

Note that the example does have multiple artifacts specified.

ITE/6/README.md
MarkLodato: Clarified.

ITE/6/README.md
MarkLodato: We decided against this for now, to keep things simple. We may want to revisit this in a future iteration.

ITE/6/README.md
MarkLodato: Please take a look at the latest revision, which has changed a bit.

ITE/6/spec/sbom.md
MarkLodato: Fixed. It is now SPDX, which is the specific format. (SBoM is not a specific format.)

ITE/6/README.md
MarkLodato: Clarified: "set of artifacts".

ITE/6/README.md
MarkLodato: No. This is now stated explicitly.

ITE/6/README.md
MarkLodato: Clarified. PTAL.

ITE/6/README.md
MarkLodato: Please take a look at the latest design.

1. `materials` is no longer in the Statement layer. Instead, it is only in the Provenance-type predicate. This means that we can more easily change it in the future. The downside means that you can't traverse the graph without understanding the Predicate type. However, the value of being able to traverse the graph without actually understanding the edge meanings is dubious. I can believe that perhaps there is a use case out there, but until we have a concrete one, our plan is to push as much down into the Predicate layer as possible. This allows each Predicate type to do what is natural in that domain. Also, things like SPDX have their own language for links, and our model now supports that.

2. Within Provenance, `materials` is an array instead of a map, so (1) the reference is by index instead of name, so there is less duplication, and (2) we have richer objects so you can attach more properties to them.

I don't follow the connection to Rekor. The big thing is standardizing `subject` so that you know what artifact the attestation is about. The links would be processed by the consumer, and they'd do further queries to walk the graph.

ITE/6/README.md
MarkLodato: I think this is addressed in the latest design (`"definedInMaterial: 0"`).

MarkLodato:
Status update: The latest version of this PR is a good candidate for v1. **Please comment on it if you have any questions or concerns.**

Our intention is to keep v1 a "minimally viable product" with only the things we are confident about supporting long term. For example, v1 does not yet support artifacts that are not identified by digest (e.g. SVN revisions). That is a feature we will need to support, but since we're not yet sure about how that will work, we will defer it to a future version.

Going forward, my plan is to move this spec to the new https://github.com/in-toto/attestation repo, where we can independently version the spec and add more docs, examples, a reference implementation, and so on. Once that happens, I'll update this ITE to effectively say "in-toto will support this new attestation format". (This is how we handled ITE-5.)

ITE/6/README.md
```diff
@@ -0,0 +1,743 @@
...
+satisfies. Type URIs are not registered; the natural namespacing of URIs is
+sufficient to prevent collisions.
+
+This ITE defines the following predicate types:
+
+*   [Provenance]: To describe the origins of a software artifact.
+*   [Link]: For migration from [in-toto 0.9].
+*   [SPDX](spec/spdx.md): A Software Package Data Exchange document.
+
+We recommend the following convetions for predicates:
```
tiziano88, line 153:
```suggestion
We recommend the following conventions for predicates:
```

ITE/6/README.md
```diff
@@ -0,0 +1,743 @@
...
+    "repo_url": "https://github.com/my-company/submodule2",
+    "repo_branch": "master",
+    "code_reviewed": true
+  }
+}
+```
+
+```json
+{
+  "attestation_type": "https://example.com/VulnerabilityScan/v1",
```
tiziano88, line 515: As an aside, I worry about this sort of attestations (e.g. vulnerability scans), since they seem to break monotonicity. Is this something that we should worry about?

ITE/6/README.md
tiziano88: Thanks Mark!

I think there is some value in representing this as a graph that can be traversed without understanding the meaning of the edges; e.g. in Oak we may want a way to recursively download and store all the attestations starting from a given artifact, and spanning the graph (perhaps up to a certain depth). This would form a sort of collection of facts about the binary, as well as its dependencies, and so on, which may be stored offline. Then a policy engine may take this as input and decide whether or not to accept the given binary (e.g. whether to run it or not).

Note that the component that does the downloading / traversal may not know anything at all about the meaning of the predicates, and the policy engine may not be able to perform queries on line.

Additionally, IMO I think there is a nice symmetry in considering each material (or set of materials, if that helps) as a directed edge in a graph (from / to the subject), in which the predicate is effectively an attribute of that edge. This way the topology of the graph should be traversable (and even cacheable, as per my previous paragraph) even without having to know anything about the individual predicates. :)

ITE/6/README.md
MarkLodato: Yes, that is a concern in general, but here it is monotonic. On line 415, we check for a positive assertion that there are zero vulnerabilities. Once we define that schema for real, we will be sure to make this point clear to policy authors. We also plan to provide policy "templates" that people instantiate, and we'll design those templates to be monotonic.

That said, the situation is a bit more nuanced. There are actually two use cases, only the first of which requires monotonicity:
- Admission control: only allow "good" things. In this case, we require positive attestation of goodness, and allow even if there exists an attestation of badness.
- Remediation: prevent "bad" things on a best effort basis. In this case, we would want to know about the attestation of badness, even if the artifact was previously allowed.

A policy that encompasses both might be "allow if there exists an attestation showing zero vulnerabilities, but deny if there exists an attestation showing a high severity vulnerability."

@nenaddedic FYI

ITE/6/README.md
MarkLodato: I can see the prospects of this idea, but it's not for v1. :-) We're looking to release v1 within the next two weeks, and there are just too many unknowns to make the links part of the Statement layer.

Our bar for putting something in the Statement layer is that there must be widespread, concrete use cases showing the need to have the field be common for all attestations. Right now, the use cases for links in the Statement layer are still theoretical.

Some challenges:

- How do you terminate the graph traversal? The graph in reality is intractably large. For example, every build likely uses 10+ artifacts, each of which uses 10+, and so on, to a really large depth. Just terminating by depth is likely not the right choice, and even then, it is a huge number of attestations. You really do need to understand the predicates to do that effectively. Alternatively, you need to figure out a model that doesn't require this deep traversal (which is what I'm currently thinking for SLSA.)

- What is the model? Right now we have Statement = Subject + Predicate. Though simple, it is easy to reason about and matches natural language. If we instead put all the graph edges in the Statement layer, we'd likely need to use a different analogy. I'm not sure that a property graph (which is what you describe) is really the right model - it doesn't easily express Provenance, for examlpe. Instead it is better represented as a hypergraph or a bipartite graph (artifacts + attestations). I **do** want to express the graph nature in the model, but it will take much more time than we have available for v1.

- Will this support all predicates easily? For example, in the current model we can simply use SPDX as a predicate and it works (though not particularly cleanly). With the links in the Statement layer, now SPDX would have to simply ignore those links (making the links feature useless) or we'd need to figure out how to rework SPDX to fit the new model (complicating onboarding).

To be clear, I share your gut feeling that it's the "right" design choice, but for v1 we'll leave it to the Predicate and see what type of conventions emerge. Once we have real-world data, we can reevaluate.

ITE/6/spec/provenance.md
```diff
@@ -0,0 +1,191 @@
...
+
+<a id="recipe.type"></a>
+`recipe.type` _string ([TypeURI]), required_
+
+> URI indicating what type of recipe was performed. It determines the meaning of
+> `recipe.entryPoint`, `recipe.arguments`, `recipe.reproducibility`, and
+> `materials`.
+
+<a id="recipe.definedInMaterial"></a>
+`recipe.definedInMaterial` _string, optional_
```
kommendorkapten, line 95: This should be int, right? It's mentioned as an int above and in the example.

MarkLodato: Oops! Fixed.

ITE/6/README.md
MarkLodato: For the record, I filed a dedicated issue for this: https://github.com/in-toto/attestation/issues/6. Let's move any further discussion to that issue.

MarkLodato:
Status update: I just tagged v0.1.0 of https://github.com/in-toto/attestation. This is a tagged release for anyone wanting to build off a non-moving target. Next step is to update this ITE to reference that repo. I hope to be able to do that next week.

PR #17, opened 2021-04-28
ITE-5: use Media Type for payloadType

MarkLodato:
Media Type is the conventional way to indicate the content type of a
payload. ITE-6 now also uses a JSON object with "_type" that indicates
its version. Therefore, this one media type can be used for all in-toto
objects going forward.


MarkLodato:
@SantiagoTorres as discussed offline.

@kommendorkapten FYI since this affects https://github.com/in-toto/in-toto-golang/pull/101

kommendorkapten:
Thanks for the update @MarkLodato. I'll address it as soon as I get time, hopefully this week.

adityasaky (approved):

PR #20, opened 2021-06-24
More signing spec -> DSSE renames

adityasaky:
I think the earlier PR only caught the ones listed as `signing-spec` and missed the non hyphenated ones.


adityasaky:
This PR also bumps the recommended version to 1.0.

MarkLodato (approved):
Oops, thanks for catching this!

PR #32, opened 2022-03-04
ITE-9: Introducing new in-toto Attestation types

adityasaky:


marcelamelara:
I in general really like the idea behind this ITE to establish a process for registering/adding new attestation types. I have a couple questions/comments about this ITE in addition to what was discussed during the community meeting today:

- What does it mean to have each recorded attestation type/predicate be its own mutable ITE? In my view new attestation types themselves aren't _really_ an extension to the in-toto framework; what seem to be the extensions are (1) the ability to support different attestation types (per ITE-6) and (2) the process for officially new (potentially blessed) attestation types (this ITE-9). 

I want to echo the comment at today's meeting that ITEs don't necessarily seem like the right/best framework for attestation object specifications. My recommendation would be to focus ITE-9 on the process of registering new attestation types, including registering new versions and maybe guidelines for specifying change notes in the attestation object repo. But it strikes me that the actual tracking of changes or updates and versioning of attestation types should be independent of in-toto as a whole (i.e., not an ITE), and rather be information provided as part of each entry in the attestation type repo (akin to how each package on PyPI has its own developer-provided landing page).

-  I assume there may be many more attestation formats that folks use with in-toto than the attestation types that are officially part of the attestation repo introduced, I'm wondering if ITE-9 should include a short list of possible benefits of going through this "official" process (e.g., peer review of attestations, avoiding too much duplication etc), to encourage folks to follow ITE-9.

- One more detail to mention in this ITE should be what the acceptable attestation data formats are. For now, it seems like the community is ok with JSON, though it might be interesting to think about what it would mean to support other data formats.

Definitely interested in folks' input on my comments! Thanks!

adityasaky:
Added in https://github.com/in-toto/ITE/pull/32/commits/8d60f901716e1c500a77070c45f70fa118f47ca0.

I personally think this ITE is in a position to consider merging as a draft, and we can evaluate the process further with some example attestation types.

trishankatdatadog:
Happy to review, lemme know when stable, thx

PR #35, opened 2022-12-23
Add old SSL signature wrapper to ITE-5 appendix

adityasaky:
By moving the old signature wrapper to the appendix of ITE-5, we can entirely remove it from the main in-toto specification.


SantiagoTorres (approved):

adityasaky:
Added a changelog entry per #36 as ITE-5 has already been accepted.

lukpueh (approved):

PR #29, opened 2021-09-16
ITE-8 Proposal Draft: provide an in-toto-library

Dentrax:
This ITE proposes creating a new repository called `in-toto-library` under `in-toto` organization. This would allow
storing _CUE_ and _OPA_ policies under repository. This policies enforce correctness and truth of [Attestation](https://github.com/in-toto/attestation) formats against pre-defined policies by running data format validating tools.

This is our first proposal on the in-toto community. 🎉  Sorry in advance for the any grammar errors. We are waiting your feedback! So we can improve the proposal according to your reviews. Still a lot to learn! 🤗 

PTAL @dlorenc @verdverm

cc: @developer-guy @erkanzileli

PR #38, opened 2023-01-10
Add ITE-10 adding policy definitions in layouts for in-toto attestations

adityasaky:
ITE-6 introduced the in-toto attestation framework. Since then, we've seen a number of attestation types be proposed and used such as SLSA provenance, runtime trace attestations, SCAI, and more. However, ITE-6 did not update in-toto layouts with mechanisms to verify the contents of new attestation types.

The latest draft of the ITE allows for verifying attribute claims at each step using lightweight DSLs built for such validations via `expectedAttributes` alongside the standard `expectedMaterials` and `expectedProducts`. It also adds similar support to current inspections. Finally, it adds a variation of a standard inspection that collects all claims pertaining to one or more subjects and applies specified constraints against those claims.

### DIFF
--- a/ITE/10/README.adoc
+++ b/ITE/10/README.adoc
@@ -164,7 +164,7 @@ apply to attestation subjects.
 
 ```yaml
 subject: string
-expectedPredicates: list of ExpectedSubectPredicates
+expectedPredicates: list of ExpectedSubjectPredicates
 ```
 
 The `subject` can be a pattern that is applied against all attestations in the
--- a/ITE/10/README.adoc
+++ b/ITE/10/README.adoc
@@ -37,7 +37,7 @@ endif::[]
 == Abstract
 
 A previous in-toto Enhancement, ITE-6, presented the in-toto Attestation
-framework. This framework introduced the idea of context-specific attestations
+Framework. This framework introduced the idea of context-specific attestations
 with their own definitions. This ITE builds on ITE-6 and proposes several
 modifications to in-toto layouts that allow for defining and verifying policies
 over ITE-6 attestations.
@@ -45,11 +45,9 @@ over ITE-6 attestations.
 [[specification]]
 == Specification
 
-=== Changes to Metadata Schema
-
 The in-toto specification describes the following schema for layouts.
 
-```
+```json
 {
     "_type": "layout",
     "expires": "<EXPIRES>",
@@ -62,16 +60,32 @@ The in-toto specification describes the following schema for layouts.
 }
 ```
 
-As a whole, in-toto breaks down every software supply chain into a number of
-steps. The specification's verification workflow uses in-toto link metadata for
-each step to ensure it was performed correctly. Each step has the following
-schema.
+An in-toto layout provides two semantics for software supply chain owners to
+define policies: steps and inspections. Steps are used to reason about and
+express constraints regarding operations _already_ performed as part of the
+supply chain. The verification is focused on ensuring authorized functionaries
+performed each step and validating the integrity of artifacts as they move from
+one step to the next. Inspections are executed _during_ verification to perform
+any other checks that are defined as external scripts.
 
-```
+TODO: this pass does not revise artifact rules for steps.
+
+This ITE primarily updates the schema and capabilities for inspections. As it
+stands, the inspections semantic can already be used to apply arbitrary
+constraints to data captured in in-toto links. Similarly, they can apply
+constraints to attestations as well. We update in-toto inspections with more
+capabilities so that they do not necessarily have to be executed as separate
+scripts. This is achieved by providing multiple options for the domain specific
+language used to express constraints. Currently, inspections only support using
+in-toto's artifact rules. With this ITE, this will remain an option, but
+inspections can also be written to not execute separate shell scripts and use
+artifact rules, and instead express data constraints using DSLs like Rego or
+Cue. The current inspection schema is as follows:
+
+```json
 {
-    "_type": "step",
+    "_type": "inspection",
     "name": "<NAME>",
-    "threshold": "<THRESHOLD>",
     "expected_materials": [
         [ "<ARTIFACT_RULE>" ],
         "..."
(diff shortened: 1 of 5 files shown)


adityasaky:
Some tangential thoughts:

* I think this will revive #29 as a way to define a library of policies for desirable properties, perhaps curated via SLSA requirements. I'll go so far as to say that this ITE's acceptance should be predicated on having a usable set of policies ready to go to prove its capabilities.

* There's probably another related ITE regarding `accepted_attestation_types` for which attestations can be used to attest to some property (or artifacts). There is, I think, overlap with discussions around predicate dictionaries and what not.

colek42:
I would like a way to write policy over a [attestation bundle](https://github.com/in-toto/attestation/blob/main/spec/bundle.md).  This would also let us create a policy linter.  @mikhailswift has some thoughts here as well.

```
```json
{
    "_type": "inspection",
    "name": "<NAME>",
    "functionaries": ["<KEYID>", "..."],
    "threshold": "<THRESHOLD>",
    "subject": ["<SUBJECT>", "..."],
    "attestations": [{
        "payloadType:" "<PAYLOADTYPE>,
        "policyLanguage": "<POLICY_LANGUAGE>",
        "policy": "<POLICY>"
         }...]
    "run": "<COMMAND>"
}
```

colek42:
Also, we can probably remove "run" and have in-toto generate an SLSA attestation.  You can then use the policy language to evaluate if the "run" command meets policy.

znewman01:
My main question here is one of motivation. I agree fully that Rego/CUE against attestations is a useful primitive. But why does it need first-class support from in-toto? You could always run an external program, so this doesn't add any *theoretical capabilities*, right?

I almost wonder if you could achieve the same thing with a small wrapper program that you could stick in `run`. Then you don't really need any in-toto changes, right?

(I might be missing something obvious here.)

znewman01:
My main question here is one of motivation. I agree fully that Rego/CUE against attestations is a useful primitive. But why does it need first-class support from in-toto? You could always run an external program, so this doesn't add any *theoretical capabilities*, right?

I almost wonder if you could achieve the same thing with a small wrapper program that you could stick in `run`. Then you don't really need any in-toto changes, right?

(I might be missing something obvious here.)

znewman01:
My main question here is one of motivation. I agree fully that Rego/CUE against attestations is a useful primitive. But why does it need first-class support from in-toto? You could always run an external program, so this doesn't add any *theoretical capabilities*, right?

I almost wonder if you could achieve the same thing with a small wrapper program that you could stick in `run`. Then you don't really need any in-toto changes, right?

(I might be missing something obvious here.)

colek42:
@znewman01

1:  in-toto policy is signed, embedding the rego/cue within the policy lets the system assert that the policy came from a trusted source.

2:  We want to improve the UX around in-toto policy evaluation.  Forcing the user to build custom workflows and tools will lower adoption rates.

3:  Policy distribution is difficult enough, having a single source of truth for what is allowed/disallowed will help simplify the process.

If you want to see an example of what this looks like in practice please take a look at my talk here:  https://www.youtube.com/watch?v=zCznRGlsYvo  and my blog article here: https://www.testifysec.com/blog/attestations-with-witness/

ITE/10/README.adoc
```diff
@@ -0,0 +1,288 @@
...
+separate script. Here, `subject` is used to identify the sets of attestations
+the policy is applied against. Further, the `functionaries` and `threshold`
+fields can be used to directly check the source of the claims being verified by
+the policy. In the current model, as inspections reason about link metadata
+generated for the execution of the inspection itself, any inspection that
+used a script to check contents of link metadata from some prior step also
+needed to verify the functionary signing the metadata in the script--in-toto's
+semantics are not used directly.
+
+TODO: the last bit of the paragraph is very convoluted. What I'm trying to say
```
trishankatdatadog, line 153: I guess you wanted to remove this, or at least merge with the paragraph above?

ITE/10/README.adoc
```diff
@@ -0,0 +1,288 @@
...
+[[reasoning]]
+== Reasoning
+
+The intent of ITE-6 and the in-toto Attestation framework is to enable software
+producers to emit a wide range of contextual information pertaining to their
+supply chain. The intent of this ITE is to extend ITE-6 with policy definitions
+such that attestation producers can validate the information in ITE-6
+attestations. It is clear from ITE-6 that there is not a one-size-fits-all
+solution to policy definitions. Therefore, to support a wide variety of use
+cases, this ITE proposes the use of existing policy languages such as Rego and
```
trishankatdatadog, line 228: One of my biggest reservations is that often these policy languages don't have cryptographic libraries baked into the language yet AFAIK.

ITE/10/README.adoc
```diff
@@ -0,0 +1,288 @@
...
+derivatives for specifying and validating policies. Policies, at their core, are
+logic programs that are applied against a set of claims, expressed here as
+in-toto attestations. This ITE uses languages such as Rego and CUE as examples
+of such languages, but leaves the specific choices to the in-toto community and
+adopters. The new layout schema proposed above includes a `<POLICY_LANGUAGE>`
+attribute for each policy to enable this variety in choice.
+
+=== Validating Policies
+
+The in-toto verification workflow is updated to support different types of
```
trishankatdatadog, line 200: How should a verifier call out to these policy languages? Seems completely unspecified to me, especially error handling.

trishankatdatadog (changes requested):

ITE/10/README.adoc
```diff
@@ -0,0 +1,288 @@
...
+    "functionaries": ["<KEYID>", "..."],
+    "threshold": "<THRESHOLD>",
+    "subject": ["<SUBJECT>", "..."],
+    "policyLanguage": "<POLICY_LANGUAGE>",
+    "policy": "<POLICY>",
+    "run": "<COMMAND>"
+}
+```
+
+Currently supported inspections _always_ execute a command or script and
```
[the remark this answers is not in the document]
adityasaky: Yes, and that's the current state of an inspection give or take. Currently, inspections are invoked using the `run` field in the layout. This may be a direct command such as a decompress as seen in in-toto/demo or the execution of a script shipped alongside the layout.

ITE/10/README.adoc
```diff
@@ -0,0 +1,288 @@
...
+    "name": "<NAME>",
+    "functionaries": ["<KEYID>", "..."],
+    "threshold": "<THRESHOLD>",
+    "subject": ["<SUBJECT>", "..."],
+    "policyLanguage": "<POLICY_LANGUAGE>",
+    "policy": "<POLICY>",
+}
+```
+
+Only `run` is removed as in this example, an external command or script is not
```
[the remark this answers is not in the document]
adityasaky: The claim above is discussing the current state of in-toto inspections.

ITE/10/README.adoc
adityasaky: That's an interesting thought I hadn't considered. My thought process with this addition was that the users could specify constraints that would apply to claims in attestations. Would policies need cryptographic capabilities? Could you help me out with an example?

znewman01: I think the point is that "@adityasaky said 'the artifact with sha256 is cool'" is hard to express in Rego.

That's solvable in two ways: first, extend your policy language to support cryptography (e.g., make it an *authorization logic*). Second, you can just do the signature validation before passing data to the policy language. So the "facts" available to Rego would all be qualified with a "so-and-so says"

ITE/10/README.adoc
znewman01: I think underspecifying is bad too. How will the attestations be made available to the policy environment? That needs to be documented. If the ITE doesn't do that, we should at least say where it will happen

ITE/10/README.adoc
adityasaky: The verification workflow essentially goes with the second option. Signatures and such are verified prior to policies being applied to claims in attestations. I'll look through this document again tomorrow to see if I can make that clearer.

adityasaky:
Pushed a third draft that revisits several ideas. With this one, we support chaining predicates that don't fit the link-esque model but are more claims / attributes oriented (eg: test run results, reviews etc). It still opens up support for embedding other DSLs throughout so attributes can be verified without relying on external scripts. It also supports the previous draft's notions about inspecting all claims (from all applicable attestations presented for verification) for one or more subjects.

JustinCappos (commented):
I wonder if it would be useful to add some examples here to motivate and ground this.  Reading the other reviewer comments, I do think this may help to address some of them as well by making this more concrete.

ITE/10/README.adoc
```diff
@@ -0,0 +1,265 @@
...
+=== Classifying Predicates
+
+Prior to ITE-6, all steps were validated against in-toto links. Inspection
+executions also resulted in the generation of in-toto links that were then
+validated. With the introduction of the in-toto Attestation Framework, it is
+necessary to rework the layout to support verifying a variety of _predicates_.
+As the framework is designed to generate claims pertaining to how software is
+produced, each predicate type can be mapped to a step in the software supply
+chain. However, it is first necessary to understand the nature of each step, and
+therefore its predicate type. in-toto was originally designed with the
```
marcelamelara, line 77: This seems to imply there is always a 1:1 correspondence between steps and predicate types. I'm not sure this is strictly true, as it's possible for a single step to generate multiple attribute type predicates along with a link type predicate. For example, a build step may simultaneously generate SLSA provenance, a runtime trace predicate about the build system, and a SCAI predicate about the build platform. All three of these predicates are tied to a particular build run, so generating them disjointly wouldn't necessarily give the expected results.

So, I might amend this description to be less restrictive about how many or which classes of predicates can be generated by a particular step.

ITE/10/README.adoc
```diff
@@ -0,0 +1,265 @@
...
+assumption that a step in the software supply chain affects in some manner the
+artifacts that constitute the supply chain. ITE-6 showed that some steps do not
+fit the materials and products model. This is a factor in _verifying_ such
+predicates using in-toto layouts.
+
+Therefore, all predicates supported for verification in an in-toto layout must
+categorized as one of two types:
```
marcelamelara, line 84: I think the term "type" is a little overloaded. We already refer to specific predicate schemas as a `predicateType`. Could we use the term "predicate class" or something like this to denote a larger grouping?

ITE/10/README.adoc
```diff
@@ -0,0 +1,265 @@
...
+
+* the link type: these are based on classic in-toto links for steps that affect
```
marcelamelara, line 86: I wonder if we can be a little more precise: instead of saying "affect", could we say "transform" or "mutate"?

ITE/10/README.adoc
```diff
@@ -0,0 +1,265 @@
...
+
+After verifying the layout and attestation signatures (including thresholds
+where applicable), attestations must be parsed to identify the materials and
+products of the step they represent. In link type predicates, the materials are
+found within the predicate and products are recorded as the attestation's
+subject. Note that alongside these artifacts, other predicate fields excluding
+the materials must also be parsed in as part of the products to enable policies
+on their contents.  For attribute type predicates, the materials are the
+attestation's subject and products are exclusively the attributes recorded
+within the predicate.
```
marcelamelara, line 154: I feel like something is missing here. A couple questions: How do verifiers map predicate types in the `predicates` field to whether they are link class or attribute class? Also, how is the `expectedAttributes` used? This description only talks about how materials and product rules may be applied in the attribute class case.

marcelamelara (commented):
Thanks @adityasaky ! A few comments on clarity and terminology.

ITE/10/README.adoc
trishankatdatadog: Right, because I don't think it's actually true: I remember Aditya pointing out that a step could use an attestation that fits one of many predicates (but here, the technical challenge is figuring out whether the [data types](https://en.wikipedia.org/wiki/Unification_(computer_science)#Application:_type_inference) are compatible, which is a fairly standard problem in computer science, but will add a lot of complexity to any in-toto verifier).

marcelamelara: I'm not sure I follow... what about my comment evoked a relationship to type inference?

trishankatdatadog: Oh, right, sorry, this is not strictly the predicate _data type_, but rather whether it's the modify-materials-products or the opposite (e.g., source code review) model. We really should use a very different term than "predicate type" for this. Nevertheless, my comment remains for multiple predicate data types per step, which I've advised Aditya about.

adityasaky: I do have a draft of steps supporting multiple predicates with policies for each. I'll sketch it out some more and update the document.

ITE/10/README.adoc
adityasaky: Also added some details to the verification workflow though I think it's not going to yet answer everything.

ITE/10/README.adoc
adityasaky: I've updated it with support for multiple predicates per step.

---

As such, this ITE proposes the following schema for step declarations.

```yaml
name: string
command: string
expectedPredicates: list of ExpectedStepPredicates
```

Each `ExpectedStepPredicate` object has the following schema.

```yaml
predicateTypes: list of TypeURIs
expectedMaterials: list of constraints
expectedProducts: list of constraints
expectedAttributes: list of constraints
functionaries: list of strings
threshold: int
```

ITE/10/README.adoc
adityasaky: > As with the current in-toto specification, the verification workflow is
presented with a root layout, public keys to verify the layout's signatures, and
a bundle of in-toto attestations. As the verifier expects a certain set of
predicate types, it is also expected to be aware of the class of each predicate.

Is this a reasonable expectation?

adityasaky:
I've split this ITE into two separate ones (with possibly a third in the works). They're at #49 and #50. I'm closing this PR in favour of those, thanks for the comments everyone!

PR #43, opened 2023-02-06
rewording suggestions

jhdalek55:
I proofread the text for clarity and readability. I made a few suggested changes, but also have the following questions:

In line 56 --is it correct to say "An attestation is an authenticated metadata" as opposed to a  "piece of metadata"? It feels like there is an agreement issue here.

In line 83, is "inspecting" the right word? 

In line 87, "must use to parse the contents of a predicate," perhaps?

### DIFF
--- a/ITE/6/README.adoc
+++ b/ITE/6/README.adoc
@@ -39,10 +39,9 @@ endif::[]
 This link:https://github.com/in-toto/ITE[in-toto enhancement (ITE)] introduces
 the in-toto Attestation Framework, a specification for generating and expressing
 verifiable claims about software supply chains. An attestation is authenticated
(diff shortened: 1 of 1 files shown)


adityasaky:
> In line 56 --is it correct to say "An attestation is an authenticated metadata" as opposed to a "piece of metadata"? It feels like there is an agreement issue here.

Updated to say "piece of authenticated metadata"

> In line 83, is "inspecting" the right word?

Switched to "using the `_type` field."

> In line 87, "must use to parse the contents of a predicate," perhaps?

Updated to indicate we mean the field specifically.

jhdalek55:
Sounds good. Otherwise it looks fine to me.

adityasaky:
closing in favour of #44

PR #46, opened 2023-04-02
ITE-6 clarifications

adityasaky:
* Update security section to rephrase each predicate type must be evaluated distinctly
* Replace "attestation type" with "predicate type" where appropriate
* Update Statement in example to v1

### DIFF
--- a/ITE/6/README.adoc
+++ b/ITE/6/README.adoc
@@ -43,8 +43,8 @@ metadata about software artifacts and the processes that generate them. The
 framework allows users to define custom "predicates" with arbitrary,
 context-specific schemas. Several predicate types covering common cases have
 already been introduced, and accompanying ITEs have been proposed to define
-processes for creating attestation types and updating in-toto layouts to verify
-them.
+processes for creating predicate types and with new in-toto layout schemas that
+can verify them.
 
 [[specification]]
 == Specification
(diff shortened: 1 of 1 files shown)


ITE/6/README.adoc
```diff
@@ -185,26 +185,26 @@ the official roadmap.
 [[security]]
 == Security
 
-This ITE does not affect the security of in-toto because:
+This ITE impacts the security of in-toto in a way that cannot be clearly defined
+here because security must be evaluated in the context of each individual
+predicate type. This is out of scope of this ITE and is delegated to each type.
```
JustinCappos, line 191:
```suggestion

Consider also that each additional predicate type supported adds complexity to implementations and 
thus adds some amount of residual risk due to the increased attack surface.  Given that signatures are 
verified before predicate type parsing is performed, this risk increase should be from trusted parties 
who are acting malicious (e.g., insider attacks).  Security weaknesses due to a predicate type are thus 
likely to result in a malicious functionary or sublayout creator being able to mask malicious behavior or 
escalate the trust in their role.

```

JustinCappos (commented):
I tried to clarify more in the security section.

JustinCappos (approved):

SantiagoTorres (approved):

PR #49, opened 2023-05-25
ITE-10: Support contextual predicates in layouts

adityasaky:
This was previously part of #38. I've broken that one down into separate ITEs so it's easier to reason about the changes proposed.

Prototyping here: https://github.com/adityasaky/in-toto-attestation-verifier

### DIFF
--- a/ITE/10/README.adoc
+++ b/ITE/10/README.adoc
@@ -1,0 +1,294 @@
+= ITE-10: Supporting Contextual in-toto Attestations in Layouts
+:source-highlighter: pygments
+:toc: preamble
+:toclevels: 2
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+.Metadata
+[cols="2"]
+|===
+| ITE
+| 10
+
+| Title
+| Supporting Contextual in-toto Attestations in Layouts
+
+| Sponsor
+| link:https://github.com/adityasaky[Aditya Sirish A Yelgundhalli]
+
+| Status
+| Draft
+
+| Type
+| Standards
+
+| Created
+| 2023-01-07
+
+|===
+
+[[abstract]]
+== Abstract
+
+A previous in-toto Enhancement, ITE-6, presented the in-toto Attestation
+Framework. This framework introduced the idea of context-specific "predicates"
+with their own definitions. This ITE builds on ITE-6 and describes how in-toto
+layouts and artifact rules can support contextual predicates.
+
+[[motivation]]
+== Motivation
+
+in-toto v1.0 defines only one type of metadata that is captured during supply
+chain operations. As such, it has a uniform verification workflow and provides
+supply chain owners the semantics, i.e. artifact rules, to verify the flow of
+artifacts in their supply chains. ITE-6 expanded the metadata formats to
+support the generation of custom attestations that have context-specific
+schemas. This means that in-toto v1.0 layouts and artifact rules cannot be used
+to verify artifacts recorded in ITE-6 attestations.
+
+[[specification]]
+== Specification
+
+Note: this document assumes the reader is familiar with ITE-6 and the in-toto
+Attestation Framework.
+
+To support different predicate types in an in-toto layout's steps, it is first
+necessary to recognize that there are two broad classes of predicates.
+
+=== Classifying Predicates
+
+As the in-toto Attestation Framework is designed to generate claims pertaining
+to how software is produced, each software supply chain step can be associated
+with one or more predicate types. in-toto v1.0 assumes that each step affects
+the artifacts that constitute the supply chain in some manner. These artifacts
+are represented as the materials and products of supply chain steps. ITE-6
+showed that some steps do not fit the materials and products model. This ITE
+shows that all predicates supported for verification in an in-toto layout can be
+categorized as one of two classes:
+
+* transformational predicates: these are based on classic in-toto links for
+  steps that transform artifacts in some manner. They consume some artifacts as
+  materials and produce others as products. The subject of such attestations
+  indicate the products of the step. Some examples of transformational
+  predicates are in-toto links and SLSA Provenance.
+* informational predicates: these predicates record contextual attributes about
+  one or more artifacts. The operations they represent consume artifacts as
+  materials (recorded as the subject of such attestations) and produce
+  attributes about them as the outcome. Examples of informational predicates are
+  test results, code reviews, and SCAI.
+
+Note that transformational predicates can also record attribute information.
+For example, the environment field in in-toto links can include such
+information. The impact of this classification is seen in how the subject of
+an attestation is handled.
+
+With this representation of different predicate types, step declarations in
+layouts can _chain_ different operations represented by distinct predicate
+types together, a fundamental feature of in-toto. For example, consider a
+supply chain with a code review step represented by a review predicate type
+that follows a source checkout step that fetches source code represented by an
+in-toto link. The review step in the layout can use in-toto's artifact rules to
+ensure the materials for the review match the products fetched by the source
+checkout step.
+
+=== Setting Predicate Type Expectations for Steps and Inspections
+
+Currently, step declarations have the following schema.
+
+```
+name: string
+threshold: int
+expectedMaterials: list of artifact rules
+expectedProducts: list of artifact rules
+pubkeys: list of authorized key IDs
+expectedCommand: list of strings
+```
+
+This ITE proposes updating this schema to incorporate predicate types.
+
+```
+name: string
+command: list of strings
+expectedMaterials: list of artifact rules
+expectedProducts: list of artifact rules
+expectedPredicates: list of expectedStepPredicates
+```
+
+Each `expectedStepPredicate` object has the following schema.
+
+```
+predicateType: string
+functionaries: list of authorized functionary key IDs
+threshold: int
+```
+
+The `predicateType` field contains the matching `TypeURI` of the predicate
+expected for the step. The original `pubkeys` field is updated to
+`functionaries` to generalize signing semantics. Finally, `threshold` applies
+per predicate type rather than the step as a whole. Note that this schema
+supports the in-toto Attestation Framework while retaining all the capabilities
+of in-toto 1.0 layouts.
+
+This change maintains the core model of the original in-toto layout while
+allowing for multiple predicate types to be associated for a single step. Note
+that while this schema allows for multiple predicate types, the expectation is
+that they share the same materials and products--if two attestations do not
+share the input and output artifacts, they must belong to different steps. Of
+course, as informational predicates do not specify artifact products, the
+`expectedProducts` field does not apply to them. In a step with exclusively
+informational predicates, the `expectedProducts` field may be omitted or left
+empty. In a step with a mixture of transformational and informational
+predicates, the field only applies to the transformational predicates. The
+`expectedMaterials` field functions as it currently does, and applies to all
+predicates, informational and transformational.
+
+=== Supporting Sublayouts
+
+The in-toto specification has support for sublayouts which are steps in a
+layout that have not links but a bundle of another layout and links which are
+recursively verified first. In in-toto v1.0, a sublayout is declared to the
+verification workflow by presenting the sublayout file in place of the link
+expected for the corresponding step.
+
+The updated in-toto layout specified in this ITE is similar. The step is
+declared with one or more expected predicate types and when a layout is
+encountered, the verifier recursively verifies the sublayout, generating the
+expected predicates in the process. Note that this expects the verifier to be
+capable of generating distinct predicates--the in-toto v1.0 verifier was only
+aware of links and sublayout verification therefore generated an in-toto summary
+link. This is discussed further on in <<Custom Verifiers>>.
+
+=== Verifying Artifact Rules
+
+The verification workflow in the in-toto specification matches links to use by
+the name of the step and the key ID used to sign it. This part of the workflow
+is updated to support matching on predicate types as well. In addition,
+artifact rules are updated to support
+link:https://github.com/in-toto/attestation/blob/main/spec/v1/resource_descriptor.md[ResourceDescriptor]
+defined in the in-toto Attestation Framework. The `name` field in a
+ResourceDescriptor object is used for pattern matching and the `digests` field
+is used to verify equality of two artifacts.
+
+```go
+func Verify(layoutEnvelope, layoutKeys, attestations, now) {
+    verifyLayoutSignatures(layoutEnvelope, layoutKeys)
+
+    layout := layoutEnvelope.Payload
+    verifyLayoutExpiry(layout, now)
+
+    if layout.Steps == nil && layout.Inspections == nil {
+        panic
+    }
+
+
+    for _, step := range layout.Steps {
+        stepAttestations := attestationsForStep(step.Name, attestations)
+        for _, predicate := range step.ExpectedPredicates {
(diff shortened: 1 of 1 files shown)


ITE/10/README.adoc
```diff
@@ -0,0 +1,278 @@
...
+step. `expectedProducts` is moved to being predicate specific as informational
+predicates may be generated alongside transformational predicates for the same
+step. This means that product rules do not generalize across all of the step's
+predicates. However, `expectedMaterials` is retained at the step level as all
+predicates generated for a step are expected to be for the same set of
+artifacts. `pubkeys` is updated to `functionaries` to generalize signing
+semantics. Finally, `threshold` applies per predicate type rather than the step
+as a whole.
+
+Similarly, inspections are also updated to support generating different
```
TomHennen, line 134: Perhaps I'm obtuse but is `expectedPredicates` the list of predicates that are generated?  Reading the bit above I read this as the list of predicates that need to be presented by this step.

ITE/10/README.adoc
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    for _, step := range layout.Steps {
+        stepAttestations := attestationsForStep(step.Name, attestations)
+        for _, predicate := range step.ExpectedPredicates {
+            predicateAttestations := attestationsForStepPredicate(predicate.Type, stepAttestations)
+            if predicate.Type == LAYOUT {
+                Verify(predicateAttestations[0], predicate.Functionaries, predicateAttestations[1:], now)
+            } else {
+                verifyAttestationSignatures(predicate.Functionaries, predicate.Threshold, predicateAttestations)
```
TomHennen, line 187: Will this error of a signature doesn't match or will it just remove unverified attestations from the list?

I suspect it would be better if it just filtered `predicateAttestations` to only have attestations that were verified and allow verification to continue if some signatures can't be verified.

Rationale: Perhaps new attestations get added to the bundle that aren't yet trusted by the layout, but also aren't necessary.

TomHennen (commented):
This makes total sense to me!

ITE/10/README.adoc
JustinCappos: This is an interesting observation.  We should consider the sort of problem you raise as a more general case.  

For example, I can also see situations where an organization wants to use a stricter sublayout, perhaps in a way that says that all dependencies must match some criteria.  So, certainly there, having additional information that isn't needed, shouldn't cause verification to fail.

ITE/10/README.adoc
```diff
@@ -0,0 +1,278 @@
...
+threshold: int
+```
+
+The `predicateType` field identifies the type of predicate expected for the
+step. `expectedProducts` is moved to being predicate specific as informational
+predicates may be generated alongside transformational predicates for the same
+step. This means that product rules do not generalize across all of the step's
+predicates. However, `expectedMaterials` is retained at the step level as all
+predicates generated for a step are expected to be for the same set of
+artifacts. `pubkeys` is updated to `functionaries` to generalize signing
```
marcelamelara, line 130: Maybe I'm missing something, but does leaving `expectedMaterials` at the step level mean that verifiers will expect to informational predicates to indicate the set of materials they apply to? Or is the assumption here that informational predicates will always indicate the `expectedMaterials` in the attestation subject field? 

I'm imagining a use case where my step generates some informational predicates about things like the tool or runtime system that ran during the step, such as the compiler or platform. Does the compiler that I ran during the step become an `expectedMaterial`? I think this also relates to my earlier comment about specifying the `expectedSubjects` in some cases rather than using the transformation-oriented terms "materials" and "products".

ITE/10/README.adoc
```diff
@@ -0,0 +1,278 @@
...
+
+Each `expectedInspectionPredicate` object has the following schema.
+
+```yaml
+predicateType: string
+expectedProducts: list of artifact rules
+```
+
+As the attestations generated during inspection executions are not signed,
+`functionaries` and `threshold` are omitted.
```
marcelamelara, line 152: Who consumes inspection attestations? I'd imagine some downstream/third-party verifier or auditor might deem the output of an inspection to be more trustworthy if it's signed.

marcelamelara (commented):
Thanks @adityasaky !

ITE/10/README.adoc
```diff
@@ -0,0 +1,278 @@
...
+command: list of strings
+expectedMaterials: list of artifact rules
+expectedPredicates: list of expectedStepPredicates
+```
+
+Each `expectedStepPredicate` object has the following schema.
+
+```yaml
+predicateType: string
+expectedProducts: list of artifact rules
```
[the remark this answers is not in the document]
adityasaky: As the two predicate classes show, the subject field may be a material or product to a step. For Provenance and Links, the subject is a product while for most others it's the material.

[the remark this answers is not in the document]
adityasaky: More broadly speaking, the materials / products semantics are retained for artifact rules and this ITE shows how they apply to all attestations, transformational and informational.

ITE/10/README.adoc
adityasaky: Yes, informational attestations will always list materials in their subject fields. However, this is not directly linked to the `expectedMaterials` being at the step level. That's to indicate that if a step can generate more than one attestation (of different predicate types, so not handled by the threshold), they must all have the same materials. Generally, it formalizes that two attestations with different predicate types must have the same materials to be considered part of the same step. Note that this does not mean that all attestations with the same materials must belong to a single step. I took this approach because not all predicates have products but they all do have materials.

>I'm imagining a use case where my step generates some informational predicates about things like the tool or runtime system that ran during the step, such as the compiler or platform. Does the compiler that I ran during the step become an expectedMaterial?

I think not, actually. The step generates this information about some artifact, so the attestation's subject is the artifact, right? The compiler info is within the predicate itself. So the attestation can be thought of representing a "tool-info" step?

ITE/10/README.adoc
adityasaky: in-toto's v1 verification workflow uses inspection links. I think this was basically so that artifact rules could be used to connect inspection results to other steps. I think the classic example is an unzip inspection which has a MATCH rule to ensure the contents of the archive match the products of the build step and so on. Nothing really stops an inspection link / attestation from being signed and distributed IMO, but we shouldn't roll that into the verification workflow because we don't expect the verifier to be capable of any signing in the first place (and indeed the verifier isn't really associated with an identity either).

ITE/10/README.adoc
adityasaky: Yes, `expectedPredicates` is what each step expects during verification. For inspections, this is just `predicates` because these attestations are generated at verification time (unsigned). I do wonder if we want to roll this back to what's in in-toto v1, where inspections generate links. If we allow multiple predicate types, the command executed during the inspection must be capable of generating an attestation of the specified type. This doesn't necessarily play well with the use cases we currently consider for inspections such as extracting from an archive to verify its contents etc. If we just stick to links, the verifier can continue generating them like in-toto v1 already does.

IMO, the _result_ of verification (VSA, summary link) is separate from inspections. We want an attestation for successful verification whether there's an inspection or not.

ITE/10/README.adoc
adityasaky: I'll update the workflow pseudocode but the idea is that all attestations for each step have their signatures verified. The ones that pass sig verification are retained after which a threshold check is applied. So attestations from untrusted keys will not abort verification if the threshold is met. This is also in-toto v1 behaviour.

ITE/10/README.adoc
adityasaky: I think the issue is primarily with the workflow which I'll update in the coming days. But more generally, I think this speaks to identifying whether inspections should generate custom predicates or not, period. The more I think about it, the more I wonder if we should leave inspections alone and just have them generate links, so no changes from in-toto v1.0 Let me take a pass at this.

marcelamelara: I think I'm in agreement, leaving inspections unchanged in this ITE makes a lot of sense to me.

adityasaky: Updated to not discuss changes to inspections.

ITE/10/README.adoc
```diff
@@ -0,0 +1,287 @@
...
+categorized as one of two classes:
+
+* transformational predicates: these are based on classic in-toto links for
+  steps that transform artifacts in some manner. They consume some artifacts as
+  materials and produce others as products. The subject of such attestations
+  indicate the products of the step. Eg. in-toto links, SLSA Provenance.
+* informational predicates: these predicates record contextual attributes about
+  one or more artifacts. The operations they represent consume artifacts as
+  materials (recorded as the subject of such attestations) and produce
+  attributes about them as the outcome. Eg. test results, code reviews, SCAI.
```
SantiagoTorres, line 82: not sure if you'd like to rephrase to `(e.g., test results, code reviews, or SCAI)` for readability

ITE/10/README.adoc
```diff
@@ -0,0 +1,287 @@
...
+that follows a source checkout step that fetches source code represented by an
+in-toto link. The review step in the layout can use in-toto's artifact rules to
+ensure the materials for the review match the products fetched by the source
+checkout step.
+
+=== Setting Predicate Type Expectations for Steps and Inspections
+
+Currently, step declarations have the following schema.
+
+```yaml
```
SantiagoTorres, line 102: Not super sure yaml is useful as a parameter for the syntax checker. It doesn't really matter much

SantiagoTorres (approved):
I like this, and I'm happy to merge this as draft. I'd still want to explore the security position of sublayouts a bit more on a future iteration (e.g,. when marking as active).

marcelamelara (commented):
Thanks for the updates @adityasaky . I have a few clarifying comments but really like where this ITE is going.

ITE/10/README.adoc
```diff
@@ -194,32 +205,30 @@ func Verify(layoutEnvelope, layoutKeys, attestations, now) {
...
 
-[[motivation]]
-== Motivation
+=== Custom Verifiers
+
+As noted above, in-toto v1.0's verification workflow generates an in-toto
+summary link if verification succeeds. This is a powerful semantic that allows
+features like sublayouts that allow for delegating the supply chain definition
+for a step to authorized functionaries, as well as summarized verification for
+environments with limited resources. With the introduction of in-toto
```
marcelamelara, line 221: Nit: clunky sentence "...semantic that allows features like sublayouts that allow for..."

ITE/10/README.adoc
```diff
@@ -194,32 +205,30 @@ func Verify(layoutEnvelope, layoutKeys, attestations, now) {
...
+Attestation Framework, the verification workflow can be parametrized to support
+generating specific predicates as needed. For example, the default verifier can
+generate an in-toto link, another can generate SLSA Provenance, while for the
```
marcelamelara, line 224: Nit: I understand that verifiers might generate different predicates based on what they are checking, but I'm not sure I follow why SLSA Provenance, in particular, could be the output of a verifier.

ITE/10/README.adoc
```diff
@@ -0,0 +1,287 @@
...
+verification workflow by presenting the sublayout file in place of the link
+expected for the corresponding step.
+
+The updated in-toto layout specified in this ITE is similar. The step is
+declared with an expected predicate type and when a layout is encountered, the
+verifier recursively verifies the sublayout, generating the expected predicate
+in the process. Note that this expects the verifier to be capable of generating
+distinct predicates--the in-toto v1.0 verifier was only aware of links and
+sublayout verification therefore generated an in-toto summary link. This is
+discussed further on.
```
marcelamelara, line 163: If possible, it would be helpful to forward-reference to the "Custom Verifiers" section here.

ITE/10/README.adoc
```diff
@@ -0,0 +1,287 @@
...
+for a step to authorized functionaries, as well as summarized verification for
+environments with limited resources. With the introduction of in-toto
+Attestation Framework, the verification workflow can be parametrized to support
+generating specific predicates as needed. For example, the default verifier can
+generate an in-toto link, another can generate SLSA Provenance, while for the
+summary use cases, a third verifier type can generate Verification Summary
+Attestations (VSAs).
+
+Each of these verifiers must be well defined with respect to the predicates they
+generate. Also, each verifier must perform the basic in-toto verification
```
marcelamelara, line 229: Can we be a bit more precise here? Not sure what "well defined with respect to the predicates" means. And is "basic in-toto verification" actually referring to the in-toto 1.0 verification flow?

marcelamelara (commented):
Adding some more comments to my review.

ITE/10/README.adoc
```diff
@@ -41,32 +41,35 @@ Framework. This framework introduced the idea of context-specific "predicates"
...
-chain. However, it is first necessary to understand the nature of each step, and
-therefore its predicate types. in-toto was originally designed with the
-assumption that a step in the software supply chain affects in some manner the
-artifacts that constitute the supply chain. ITE-6 showed that some steps do not
-fit the materials and products model. This is a factor in _verifying_ such
-predicates using in-toto layouts.
-
-Therefore, all predicates supported for verification in an in-toto layout must
+As the in-toto Attestation Framework is designed to generate claims pertaining
+to how software is produced, each predicate type can be mapped to a step in the
```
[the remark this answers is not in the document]
adityasaky: I flipped the clause like you suggested.

ITE/10/README.adoc
```diff
@@ -109,62 +114,64 @@ This ITE proposes updating this schema to incorporate predicate types.
...
+supports the in-toto Attestation Framework while retaining all the capabilities
+of in-toto 1.0 layouts.
+
+This change maintains the core model of the original in-toto layout while
+allowing for multiple predicate types to be associated for a single step. Note
+that while this schema allows for multiple predicate types, the expectation is
+that they share the same materials and products--if two attestations do not
+share the input and output artifacts, they must belong to different steps. Of
+course, as informational predicates do not result in products, the
+`expectedProducts` field does not apply to them.  In a step with exclusively
```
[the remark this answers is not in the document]
adityasaky: Updated to read "informational predicates do not specify artifact products"

ITE/10/README.adoc
```diff
@@ -109,62 +114,64 @@ This ITE proposes updating this schema to incorporate predicate types.
...
-expects a layout and identifies it. This layout is then verified recursively
-before verification continues at the original layout.
+The in-toto specification has support for sublayouts which are steps in a
+layout that have not links but a bundle of another layout and links which are
+recursively verified first. In in-toto v1.0, a sublayout is declared to the
+verification workflow by presenting the sublayout file in place of the link
+expected for the corresponding step.
+
+The updated in-toto layout specified in this ITE is similar. The step is
+declared with an expected predicate type and when a layout is encountered, the
```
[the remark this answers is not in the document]
adityasaky: Updated

ITE/10/README.adoc
adityasaky: In some cases, SLSA provenance can be thought of as a compilation of information across multiple steps involved in building some software. For example, there may be links prior to the actual build that record environment setup or dependency provisioning. The provenance for such a build needs to include the information from such steps as well so I think we ought to explore how in-toto verification itself can collate this information across steps. There's definitely more to do here.

ITE/10/README.adoc
adityasaky: Restated this a bit.

ITE/10/README.adoc
marcelamelara: Ok, thanks for clarifying. I can definitely see this sort of use case for SLSA, but the Provenance v1 seems a little less amenable to expressing non-build operations at first glance, so would definitely be interested in seeing some more concrete examples/suggestions for this.

marcelamelara (approved):
Thank you for all your work on this @adityasaky !

PR #50, opened 2023-05-25
ITE-11: Support attribute constraints in layouts

adityasaky:
This was previously part of #38. I've broken that one down into separate ITEs so it's easier to reason about the changes proposed.

Prototyping here: https://github.com/adityasaky/in-toto-attestation-verifier

### DIFF
--- a/ITE/11/README.adoc
+++ b/ITE/11/README.adoc
@@ -1,0 +1,291 @@
+= ITE-11: Verifying Attributes in in-toto Attestations
+:source-highlighter: pygments
+:toc: preamble
+:toclevels: 2
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+.Metadata
+[cols="2"]
+|===
+| ITE
+| 11
+
+| Title
+| Verifying Attributes in in-toto Attestations
+
+| Sponsor
+| link:https://github.com/adityasaky[Aditya Sirish A Yelgundhalli]
+
+| Status
+| Draft
+
+| Type
+| Standards
+
+| Created
+| 2023-01-07
+
+|===
+
+[[abstract]]
+== Abstract
+
+in-toto metadata, whether legacy links or contextual predicates defined under
+the Attestation Framework, can record various attributes about software supply
+chain steps and artifacts. However, layouts do not intrinsically support
+setting constraints on these attributes, passing off these capabilities to the
+"inspections" semantic instead. This ITE proposes adding support for setting
+such constraints in in-toto layouts.
+
+[[specification]]
+== Specification
+
+This ITE proposes embedding attribute verification capabilities within in-toto
+layouts. Specifically, it proposes a new step and inspection specific field
+called `expectedAttributes` that will live alongside the existing
+`expectedMaterials` and `expectedProducts` fields. Unlike those fields,
+`expectedAttributes` will not support artifact rules but will instead evaluate
+each rule using the Common Expression Language (CEL). Every rule MUST evaluate
+to true for the verification to succeed.
+
+If this ITE is applied to the in-toto v1.0 layout schema, each step would look
+as follows.
+
+```yaml
+name: string
+threshold: int
+expectedMaterials: list of artifact rules
+expectedProducts: list of artifact rules
+expectedAttributes: list of CEL expressions
+pubkeys: list of authorized key IDs
+expectedCommand: list of strings
+```
+
+If instead the ITE is applied to ITE-10 layouts with support for the in-toto
+Attestation Framework, the resultant step schema is as follows.
+
+```yaml
+name: string
+command: list of strings
+expectedMaterials: list of artifact rules
+expectedProducts: list of artifact rules
+expectedPredicates: list of expectedStepPredicates
+```
+
+Each `expectedStepPredicate` object has the following schema.
+
+```yaml
+predicateType: string
+expectedAttributes: list of CEL expressions
+functionaries: list of authorized functionary key IDs
+threshold: int
+```
+
+Similarly, inspections are updated to the following schema.
+
+```yaml
+name: string
+expectedMaterials: list of artifact rules
+expectedProducts: list of artifact rules
+expectedAttributes: list of CEL expressions
+run: list of strings
+```
+
+=== Verifying Attribute Constraints
+
+An in-toto implementation supporting these checks must support the evaluation of
+CEL expressions. During the in-toto verification workflow, after applying rules
+from `expectedMaterials` and `expectedProducts`, each rule is evaluated against
+the attestation it applies to. If the rule expression evaluates to true, the
+rule passes and verification proceeds. If the expression evaluates to false,
+verification is aborted with an error indicating the failing rule. For example,
+if a layout has two rules, `rule1` and `rule2`, verification only succeeds if
+both `rule1` AND `rule2` evaluate to true.
+
+[[motivation]]
+== Motivation
+
+in-toto attestations, whether link metadata from in-toto v1.0 or contextual
+metadata from the Attestation Framework, contain various claims. The link
+metadata format has two opaque fields, `environment` and `byproducts`, that can
+hold arbitrary information. Contextual predicates also record various software
+supply chain specific attributes, though in predefined fields rather than in an
+arbitrary manner in opaque fields. In the current in-toto specification, the
+only way to verify these attributes are through inspections that require
+executing commands or scripts during verification. If invoked via scripts, they
+need to be communicated alongside layouts to the verifier and there are no
+in-toto specific guarantees as to their integrity or if they were issued by the
+authors of the layout. Further, these scripts can be complicated as they need
+to load attestations, once again verify their signatures using keys in the
+layout, before they can actually validate the attributes in question.
+
+=== Example: Verifying Result of a Test Run
+
+The in-toto Attestation Framework has a vetted predicate capturing the results
+of
+link:https://github.com/in-toto/attestation/blob/main/spec/predicates/test-result.md[test runs].
+To verify that the test result attestation indicates a successful test run, the
+following CEL rule can be used.
+
+```yaml
+...
+expectedAttributes:
+    - "predicate.result == 'PASSED'"
+...
+```
+
+=== Example: Verifying SLSA Provenance Attributes
+
+link:https://github.com/in-toto/attestation/blob/main/spec/predicates/provenance.md[SLSA Provenance]
+records various attributes such as the identity of the builder, source
+repository, and various other parameters. These attributes can be verified in
+in-toto layouts as follows.
+
+```yaml
+...
+expectedAttributes:
+    - "predicate.runDetails.builder.id == '<expected value>'"
+    - "predicate.externalParameters.repository == '<expected value>'"
+    - "predicate.externalParameters.ref == '<expected value>'"
+...
+```
+
+=== Example: Verifying Build Isolation via Runtime Trace Attestation
+
+This vetted predicate is used to
+link:https://github.com/in-toto/attestation/blob/main/spec/predicates/runtime-trace.md[record a runtime trace]
+of some supply chain step, such as the build step. It can be used to verify
+that the builder made no network calls. But first, it is necessary to verify
+that the monitor was of an expected type and that its trace policy included
+tracking network calls.
+
+```yaml
+...
+expectedAttributes:
+    - "predicate.monitor.type == 'https://github.com/cilium/tetragon'"
+    - "predicate.monitor.tracePolicy.policies.exists(p, p['Name'] == 'connect')"
+    - "size(predicate.monitorLog.network) == 0"
+...
+```
+
+=== Example: Verifying Contents of Opaque Fields in in-toto Links
+
+As noted above, in-toto links have an opaque `byproducts` field that records
+additional information pertaining to the step performed. in-toto's verification
+workflow does not verify its contents by default. However, the changes proposed
+here can be used to add checks for its contents as well as those of the other
+opaque field, `environment`.
+
+```yaml
+...
+expectedAttributes:
+    - "size(predicate.byproducts['stderr']) == 0"
+    - "predicate.byproducts['return-value'] == 0"
+    - "predicate.environment['<key>'] == '<expected value>'"
+...
+```
+
+[[reasoning]]
+== Reasoning
+
+This ITE presents certain changes that necessitate further reasoning.
+
+=== Use of CEL
+
+CEL is a lightweight language designed for expression evaluations. One of its
+key applications is for security policies, making it a good fit for specifying
+attribute rules in in-toto layouts. CEL expressions are evaluated in linear time
+and the language is not Turing-complete, mitigating concerns about arbitrary
+executions during verification. Also, it may be possible to generate CEL rules
+from other higher-level languages to avoid having developers write CEL
+expressions manually.
+
+Finally, another feature making CEL a good fit for in-toto layout attribute
+checks is that it was designed to be embedded in other applications. Indeed,
+in-toto is not the first system to support embedded CEL expressions--Kubernetes
+also supports using CEL to declare various constraints as does Envoy in its Role
+Based Access Control policies.
+
+=== Supporting Other Languages
+
+Other policy specific languages such as Open Policy Agent's Rego and CUE are
+alternatives to using CEL. Also, other expression languages with similar
+capabilities as CEL may also be preferred by some users. Supporting these
+choices (and identifying the language used for a rule) is an open discussion
+point. CEL is simple to write and embed into in-toto layouts but is also less
+powerful than full-fledged policy languages. That said, the focus is on setting
+constraints on attributes recorded in in-toto attestations, which CEL is capable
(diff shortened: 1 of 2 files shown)


mikhailswift:
CEL seems like a solid choice for this. It's simple enough to reason about easily while still being powerful enough to do the vast majority of the types of policies surrounding predicates that we care about.

Cole and I have been working with some more complex policies that have us thinking about writing policies over multiple predicates. However, I think that idea would need to be fleshed out a lot further before we could really reason about them here and if we can avoid that I think we'd be better off for it.

ITE/11/README.adoc
```diff
@@ -0,0 +1,291 @@
...
+== Specification
+
+This ITE proposes embedding attribute verification capabilities within in-toto
+layouts. Specifically, it proposes a new step and inspection specific field
+called `expectedAttributes` that will live alongside the existing
+`expectedMaterials` and `expectedProducts` fields. Unlike those fields,
+`expectedAttributes` will not support artifact rules but will instead evaluate
+each rule using the Common Expression Language (CEL).
+
+If this ITE is applied to the original in-toto layout, the schema for each step
```
TomHennen, line 56: nit: I find the "If this ITE is applied" language a bit confusing (especially given the "If instead" construction below).

The conditionality makes it sound like it's up to the user to decide where to use the ITE, when the proposal is actually to apply the ITE to the entire spec and make all of these formulations possible.

If this just the standard way to write ITEs that's cool.

ITE/11/README.adoc
```diff
@@ -0,0 +1,291 @@
...
+that require executing commands or scripts during verification. If invoked via
+scripts, they need to be communicated alongside layouts to the verifier and
+there are no in-toto specific guarantees as to their integrity or if they were
+issued by the authors of the layout. Further, these scripts can be complicated
+as they need to load attestations, once again verify their signatures using keys
+in the layout, before they can actually validate the attributes in question.
+
+=== Example: Verifying Result of a Test Run
+
+The in-toto Attestation Framework has a vetted predicate capturing the results
```
TomHennen, line 145: optional: link to these various predicates to make this easier to follow?

TomHennen (commented):
This seems like a big improvement.  It seems like combining CEL with in-toto might resolve some of the difficulties of using CEL too.

ITE/11/README.adoc
adityasaky: The conditional is meant to apply to the type of layout. It might help to replace the original in-toto layout with references to v1.0. The reason is, this could easily apply to legacy layout schemas without depending on the layout proposed in #49.

ITE/11/README.adoc
adityasaky: Done!

trishankatdatadog (commented):
My biggest feedback for this ITE is that it will be impossible to get consensus on an expression language of choice. We could ofc be strongly opinionated about it, but I think we should repeat what we did with signing envelopes: while we do not mandate any particular one, we strongly recommend DSSE, and support it out-of-the-box. We should be similarly liberal about the choice of policy engine while choosing one by default.

trishankatdatadog (approved):
As I said earlier in a [comment](https://github.com/in-toto/ITE/pull/50#pullrequestreview-1653754386), my biggest feedback for this ITE is that it will be impossible to get consensus on an expression language of choice. We could ofc be strongly opinionated about it, but I think we should repeat what we did with signing envelopes: while we do not mandate any particular one, we strongly recommend DSSE, and support it out-of-the-box. We should be similarly liberal about the choice of policy engine while choosing one by default.

However, I am happy to approve this PR as a working draft so that we can iterate and make progress. Sorry for the delay in review, and thanks for all your hard work!

adityasaky:
I agree, I think it makes sense to generally call for an expression language with a set of properties, and recommend the use of a single one, possibly / likely CEL.

trishankatdatadog:
> I agree, I think it makes sense to generally call for an expression language with a set of properties, and recommend the use of a single one, possibly / likely CEL.

To properly generalise at the risk of slightly complicating this ITE, there may be need to be an `attributesType` similar to `predicateType`, either at the top-level (simpler but more rigid) or predicate (more complicated but more flexible) scope.

We may also want to discuss the security implications (e.g., computational or memory constraints) of choosing any expression language.

ITE/11/README.adoc
```diff
@@ -0,0 +1,278 @@
...
+expectedMaterials: list of artifact rules
+expectedPredicates: list of expectedStepPredicates
+```
+
+Each `expectedStepPredicate` object has the following schema.
+
+```yaml
+predicateType: string
+expectedProducts: list of artifact rules
+expectedAttributes: list of CEL expressions
```
MarkLodato, line 84: A raw list of CEL expressions might be a bit limiting. If something fails, it's difficult to present a clear error message of what went wrong. You might want to have a field for error messages. (Similarly, giving each an optional name might make it more readable.)

For example:

```yaml
expectedAttributes:
  - name: builderId
    require: "predicate.runDetails.builder.id == 'https://example.com/myBuilder'"
    errorMessage: '"Unexpected builder.id: expected \"https://example.com/myBuilder\", got " + predicate.runDetails.builder.id'
  - name: repository
    require: "predicate.externalParameters.repository == '<expected value>'"
    errorMessage: '...'
```

ITE/11/README.adoc
```diff
@@ -0,0 +1,278 @@
...
+
+in-toto metadata, whether legacy links or contextual predicates defined under
+the Attestation Framework, can record various attributes about software supply
+chain steps and artifacts. However, layouts do not intrinsically support
+setting constraints on these attributes, passing off these capabilities to the
+"inspections" semantic instead. This ITE proposes adding support for setting
+such constraints in in-toto layouts.
+
+[[specification]]
+== Specification
```
MarkLodato, line 47: Sorry if I'm missing context, but how do you do graph traversal? For example, if you want to have a policy that says that the artifact must have a SLSA provenance attestation where each `resolvedDependency` has its own VSA attestation having some other properties?

Is that a general feature of in-toto layouts that is orthogonal to this?

ITE/11/README.adoc
```diff
@@ -0,0 +1,278 @@
...
+
+This ITE proposes embedding attribute verification capabilities within in-toto
+layouts. Specifically, it proposes a new step and inspection specific field
+called `expectedAttributes` that will live alongside the existing
+`expectedMaterials` and `expectedProducts` fields. Unlike those fields,
+`expectedAttributes` will not support artifact rules but will instead evaluate
+each rule using the Common Expression Language (CEL).
```
MarkLodato, line 54: nit: can you state explicitly that `expectedAttributes` are AND'd together?

ITE/11/README.adoc
```diff
@@ -0,0 +1,278 @@
...
+    - "predicate.environment['<key>'] == '<expected value>'"
+...
+```
+
+[[reasoning]]
+== Reasoning
+
+This ITE presents certain changes that necessitate further reasoning.
+
+=== Use of CEL
```
MarkLodato, line 196: Might also be worth noting that users are not necessarily expected to write raw CEL themselves, but rather higher-level languages could compile down to CEL.

ITE/11/README.adoc
MarkLodato, line 84: FYI, one thing that's done in CEL Policy Templates is to create reusable `terms` for more complicated policies. Here's an example:
https://github.com/google/cel-policy-templates-go/blob/master/test/testdata/binauthz/template.yaml

Not sure if you want to add that proactively, or if you want to wait to see if it's actually a problem.

ITE/11/README.adoc
adityasaky: Somewhat, though it's not currently part of the spec. @marcelamelara has worked on this recently in the context of SCAI evidence references, to go resolve the referenced attestation for use in the verification process. I view this as possibly a preprocessor / ahead of time run, so at layout verification the required attestations are present. This may, however, become problematic with lots of references, only some of which are actually required. I'm open to fleshing this out some more, but I suspect this is a more general in-toto concern and not just for ITE-11.

Did I parse your question correctly?

ITE/11/README.adoc
adityasaky: I suspect we'll need something like this eventually but I want to wait and see what shape this takes. In part, this is because I think we need to explore how layout-templates can be written, which may be where we want to support this.

ITE/11/README.adoc
MarkLodato: Yes, you parsed it correctly. So in short, that's a known limitation and this is a first step where you can make decisions that don't require graph traversal? And that a future ITE could bring better graph traversal options?

ITE/11/README.adoc
MarkLodato: Sounds good, feel free to resolve

ITE/11/README.adoc
```diff
@@ -0,0 +1,285 @@
...
+powerful than full-fledged policy languages. That said, the focus is on setting
+constraints on attributes recorded in in-toto attestations, which CEL is capable
+enough for. This may change if use cases emerge that CEL cannot express
+constraints for. It's also worth noting that for niche scenarios, the original
+method of writing custom inspections also remains an option.
+
+[[backwards-compatibility]]
+== Backwards Compatibility
+
+This ITE is entirely backwards compatible with both the in-toto v1.0 layout
```
JustinCappos, line 231: Backwards compatible with what?  Is this a legacy client and a new layout?  A new client and a legacy layout, etc.?  I don't think it is always backwards compatible.

ITE/11/README.adoc
```diff
@@ -0,0 +1,285 @@
...
+attribute verification. This does not weaken in any way the security properties
+of such scenarios.
+
+[[security]]
+== Security
+
+There are two aspects to the security implications of this ITE. The first is the
+addition of a substantial surface--the in-toto specification is at least in part
+dependent on the CEL specification, and implementations are similarly dependent
+on the CEL engine. TODO: flesh this out some more.
```
JustinCappos, line 244:
```suggestion
on the CEL engine. Hence the security properties of the query language are paramount.
```

JustinCappos (approved):
I'm okay to have this merge in as a draft.  I would like to see the backwards compatibility part worked on and a little more of the links to the due diligence we did on CEL listed in the document.

ITE/11/README.adoc
adityasaky: Clarified this some more to indicate that backwards compatibility isn't a given, we may have situations where a check that ought to have failed actually passes because a legacy client skips attribute checks.

ITE/11/README.adoc
adityasaky: https://github.com/in-toto/ITE/issues/55


This is really great! I'm going to work on this and support it in attestation-verifier, but without blocking merging this ITE as draft as it's been an open PR for some time now.

ITE/11/README.adoc
```diff
@@ -0,0 +1,291 @@
...
+to true for the verification to succeed.
+
+If this ITE is applied to the in-toto v1.0 layout schema, each step would look
+as follows.
+
+```yaml
+name: string
+threshold: int
+expectedMaterials: list of artifact rules
+expectedProducts: list of artifact rules
```
nmiyake, line 64: Checking for clarification: [the in-toto spec for the Steps format](https://github.com/in-toto/specification/blob/master/in-toto-spec.md#431-steps) specifies these fields as `expected_materials` and `expected_products`.

I haven't seen any ITEs or spec updates that change this, but ITE-10 and this ITE-11 (as well as the project at https://github.com/in-toto/attestation-verifier) seem to use this `camelCase` for these same fields.

Is it the case that:
1. The documentation refers to these fields in `camelCase`, but that's just for convenience, and actual serialization should remain `snake_case`
2. The layout schema outlines the fields that should exist, but the actual naming/serialization format for the fields is considered implementation-dependent and is not important to standardize
3. There is an intent to move to `camelCase` for the standard naming of fields in the layout schema, but the spec just hasn't been updated

We are using our own in-toto implementation for verification and are attempting to adhere to the spec as much as possible/practical, but are having a hard time determining whether there is a correct/canonical serialization format based on these inconsistencies.

adityasaky: For me personally, it was mostly 1, though it's something to consider and nail down for compatibility / consistency. Happy to change these back to snake case unless there's interest in using this breaking change to also change the serialization format. I don't feel strongly one way or another.

FWIW, I think we should probably reject 2 if it does come up; it's tempting to treat this as an implementation detail but we should aim for maximum compatibility of layouts and corresponding tooling. @jkjell we should coordinate this, if applicable, wrt witness policies too.

ITE/11/README.adoc
```diff
@@ -0,0 +1,291 @@
...
+command: list of strings
+expectedMaterials: list of artifact rules
+expectedPredicates: list of expectedStepPredicates
+```
+
+Each `expectedStepPredicate` object has the following schema.
+
+```yaml
+predicateType: string
+expectedProducts: list of artifact rules
```
nmiyake, line 84: Is `expectedProducts` meant to be at this level, or at the step schema level? In the previous section that discusses applying this ITE to the 1.0 schema, `expectedProducts` is still at the `step` level, and that is also where it is in the `in-toto/attestation-verifier` project: https://github.com/in-toto/attestation-verifier/blob/fc1406e6ae8b41696091061712feaee6bfdd263a/verifier/models.go#L40

nmiyake:
Hi @adityasaky, I was wondering if you could comment on the overall state of ITE-10 and ITE-11 and their relations to the in-toto spec. It looks like ITE-10 merged about a year ago (with "Draft" status), and the core of the ITE-11 PR was from mid to late last year as well. The [core in-toto spec](https://github.com/in-toto/specification/blob/master/in-toto-spec.md) is at v1.0.0 as of June 2023.

What is the expected process/timeline for ITE-10 and ITE-11 to advance and/or graduate from draft status? Additionally, both ITE-10 and ITE-11 propose changes to the layout file schema, so in order for them to be "officially" used, the core spec would need to be updated to reflect them as well. Was just wondering whether there were specific tasks or actions that needed to occur for this to happen that are actively in progress, or if this is somewhat in limbo/holding pattern due to external factors.

As context, I am interested in migrating to using the ITE-10/11 version of in-toto verification with attestations, and although it is possible for me to proceed based on the ITEs, ideally it would be nice to be able to be using a release version of in-toto (rather than having to caveat that I'm implementing "v1.0.0 with ITE-10 and ITE-11 proposals as of July 2024"). Thanks!

adityasaky:
Hi @nmiyake! The general process is an ITE gets merged as draft, and after it's been implemented and we've had a chance to receive feedback from consumers, we decide whether it can be accepted. At that point, we'd update the spec to incorporate the ITE.

The status of ITE-10 and 11 is we're essentially waiting for feedback on if the new layout schemas make sense for different consumers and their use cases, especially given the magnitude of the changes. We actually just discussed getting ITE-11 merged as a draft soon, @trishankatdatadog will be helping drive that. I hope that once we get this PR merged, and we reconcile any drift in ITE-10/11 with attestation-verifier / other implementations of the ITEs, we'll start hearing (from you and others!) what else needs to go into these ITEs.

On a related note: we kicked off a working group to drive in-toto policies onwards today. We would love to have you on those calls, they should be on a public calendar or so somewhere in the near future, I think. cc @trishankatdatadog, @jkjell, @forrin

trishankatdatadog:
> We actually just discussed getting ITE-11 merged as a draft soon, @trishankatdatadog will be helping drive that.

I approved it last year, apparently, and completely forgot. @SantiagoTorres, could we please get a LGTM from you to unblock?

> On a related note: we kicked off a working group to drive in-toto policies onwards today. We would love to have you on those calls, they should be on a public calendar or so somewhere in the near future, I think. cc @trishankatdatadog, @jkjell, @Forrin

@nmiyake Just ping me on the CNCF Slack and I'll be happy to add you, thanks.

ITE/11/README.adoc
adityasaky: Good catch, for the new layout, we were experimenting with having products at the predicate level if we wanted to support multiple predicates per step, but eventually, we moved it back up. Attributes still apply at the predicate level. I'll make this change shortly.

nmiyake:
@adityasaky @trishankatdatadog thanks! I've joined the CNCF Slack and pinged there for further information.

For context, our organization has been using in-toto to perform verification of deployment artifacts since around late 2021 using the pre-1.0 version. We're in the process of migrating to use attestations so that we can record and verify more information, and as of right now are planning to use a combination of in-toto v1 + ITE-10/ITE-11 proposals.

As you can probably tell from some of my comments, in doing so I've found some occasional inconsistencies between different parts of the specs and reference implementations. From the external perspective, it's been hard to tell exactly what the status is for the specs and reference implementations etc., so getting some initial context around that would probably be most helpful for me. Although my primary focus is for our internal use case, I think that having a robust and accurate official spec and reference implementation(s)/libraries would be a benefit to everyone, and I'd be happy to look to contribute work across the spec and any Go/Java implementations, but would want to make sure that I have the proper overall context first (for example sounds like the current Go library may be deprecated in favor of go-witness, which isn't context I knew about prior to some of the comments here).

Happy to engage either via a comment thread here or over Slack or email if it makes sense to do so. Thanks!

SantiagoTorres:
Hi all.

This PR has received plenty of feedback, and I think it's come a long way. I'm gleaning from the conversations that everybody is happy to merge as *draft*.

There's still space to refine things before it's marked as accepted.

PR #52, opened 2023-10-02
ITE-9: Require predicateName for predicates

adityasaky:
The short predicateName is expected to be unique as well, and is to be used for things like the file name for the predicate as well as special media types that also indicate the predicate contained within.

In addition, this PR updates a link to the in-toto attestation spec, pointing to the formal v1.0 release instead of the older draft version.

Ref: https://github.com/in-toto/attestation/pull/283#discussion_r1336038877


ITE/9/README.adoc
```diff
@@ -45,7 +45,7 @@ The link:https://github.com/in-toto/attestation/[in-toto Attestation Framework]
...
+The specification MUST contain a detailed schema of the fields that make up the predicate. This schema MUST define unambiguously a unique `predicateType` and a shorter `predicateName` following the rules defined by the attestation framework. The `predicateName` is expected to be used as the predicate specification's file name as well as a hint in certain contexts such as a media type to indicate the predicate embedded in the attestation. The name must not include version information and is expected to only be used as an unauthenticated hint. The schema must also clearly define the types of every field, with ranges for values where applicable. Further, the specification MUST contain rules for parsing the fields, as well as make recommendations for handling edge cases, where parsing or handling of fields would deviate from the standard link:https://github.com/in-toto/attestation/tree/v1.0/spec/v1.0#parsing-rules[parsing rules]. For example, it may be preferable in some cases to silently allow extra, undefined fields, while in other cases such metadata should be rejected.
```
trishankatdatadog, line 48: Why is the `predicateName` considered unathenticated? Is it not covered by a signature?

adityasaky: This is not listed in the predicate itself. This is a change to the _specification_ of a predicate, like this document: https://github.com/in-toto/attestation/blob/main/spec/predicates/test-result.md. The proposal is to add, alongside the type URI, a field that basically says "Predicate name: `test-result`". This short name can then be used in media types and the like to provide an unauthenticated hint at the predicate type embedded in the corresponding attestation without getting into the version and other aspects. See: https://github.com/in-toto/attestation/issues/271

trishankatdatadog: I don't understand: why is it considered _unauthenticated_ when it seems to be in the signed portion of an attestation?

adityasaky: This is _not_ in the attestation, this is in the markdown file that describes a predicate.

trishankatdatadog: I see, thanks for the clarification. This is a bit confusing TBH. Are the other fields that are in the schema but not in the actual predicate? If not, why not?

adityasaky: ITE-9 is a process ITE that describes how users can submit a new predicate to be vetted. As such, it requires each predicate definition to include some specific information such as its type URI (used in the attestation statement), other fields specific to the predicate, use cases, etc. This PR merely adds an additional field, the predicate name, to that document. None of the things listed in ITE-9 go directly in a particular predicate, but rather the markdown file for each predicate in https://github.com/in-toto/attestation/tree/main/spec/predicates.

trishankatdatadog: I see. An ITE does seem overkill for this, but as long as we clearly explain that a predictName is metadata for a predicate scheme (which is already a bit confusing because the predicate is also metadata but on a lower level), then we should be good. Saying it is "an unauthenticated hint" is more confusing than illuminating because that hint isn't actually in the predicate, and it wouldn't be unathenticated if it was part of the a predicate.

adityasaky: I specified "unauthenticated hint" to set expectations on how the predicate name should be used. Let me try and tweak the phrasing.

adityasaky: Moved things around to connect unauthenticated hint and the text about using it in media types. Does that help?

My thinking was that this was fine as ITE-9 only features at the time of introduction of a predicate. Once it's vetted, the attestation + verification specs handle how and what must be verified. WDYT?

trishankatdatadog: I'm really not trying to be difficult, but I _still_ don't understand how the predicateName is going to drastically change the media type of what's in the specification file (aren't they all just [Markdown](https://github.com/in-toto/attestation/tree/main/spec/predicates) files?). I don't think I'm the best editor for this, so I'm going to defer to @marcelamelara here 😂

marcelamelara: @trishankatdatadog I haven't reviewed this PR yet. But maybe some extra context may help. A while back, we were asked, on the attestation side, to provide guidance on predicate-level media types. This is helpful in settings where consumers of attestations may need a hint about the contents of a DSSE without unpacking the entire envelope. We (the attestation maintainers) decided, for a few reasons, that a predicate name would be the most succinct way of defining unique predicate media type patterns for such use cases. The practice of choosing unique predicate names is already in place, but currently only implicitly, so this PR seeks to make that practice explicit.

ITE/9/README.adoc
```diff
@@ -45,7 +45,7 @@ The link:https://github.com/in-toto/attestation/[in-toto Attestation Framework]
...
+The specification MUST contain a detailed schema of the fields that make up the predicate. This schema MUST define unambiguously a unique `predicateType` and a shorter `predicateName` following the rules defined by the attestation framework. The `predicateName` must be used as the predicate specification's file name. The name must not include version information and is expected to only be used as an unauthenticated hint in other contexts such as a media type to indicate the predicate stored in an attestation. The schema must also clearly define the types of every field, with ranges for values where applicable. Further, the specification MUST contain rules for parsing the fields, as well as make recommendations for handling edge cases, where parsing or handling of fields would deviate from the standard link:https://github.com/in-toto/attestation/tree/v1.0/spec/v1.0#parsing-rules[parsing rules]. For example, it may be preferable in some cases to silently allow extra, undefined fields, while in other cases such metadata should be rejected.
```
marcelamelara, line 48: Nit: Can you please wrap at 80 chars?

marcelamelara: What may be misleading here is the term "schema" in the second sentence. Both the predicateType and predicate name aren't meant to be included in the predicate schema itself. Suggesting a few wording changes...

```suggestion
The specification MUST define unambiguously a unique `predicateType`, a shorter `predicateName` following the rules defined by the attestation framework, and contain a detailed schema of the fields that make up the predicate. Any files related to the predicate specification (spec document, directories, protobuf definition) MUST match the `predicateName`. The name must not include version information and is expected to only be used as an unauthenticated hint in other contexts such as a media type to indicate the predicate stored in an attestation. The schema must also clearly define the types of every field, with ranges for values where applicable. Further, the specification MUST contain rules for parsing the fields, as well as make recommendations for handling edge cases, where parsing or handling of fields would deviate from the standard link:https://github.com/in-toto/attestation/tree/v1.0/spec/v1.0#parsing-rules[parsing rules]. For example, it may be preferable in some cases to silently allow extra, undefined fields, while in other cases such metadata should be rejected.
```

marcelamelara (changes requested):
Thanks @adityasaky ! I have a few suggested edits to the text.

ITE/9/README.adoc
marcelamelara: One more comment: It may be helpful to provide an example for how `predicateName` should be used in file naming.

ITE/9/README.adoc
trishankatdatadog: I think I finally (sorta) get it. Left some comments in that PR to get more clarifications. Thanks!

ITE/9/README.adoc
adityasaky: > Nit: Can you please wrap at 80 chars?

happy to wrap this paragraph. For some reason, I don't seem to have submitted ITE-9 originally and I wasn't sure about making out-of-context changes.

adityasaky: I applied your suggestion. I also added an example for the name.

ITE/9/README.adoc
adityasaky: @trishankatdatadog can you check the updated text?

ITE/9/README.adoc
```diff
@@ -45,7 +45,26 @@ The link:https://github.com/in-toto/attestation/[in-toto Attestation Framework]
...
 
-The specification MUST contain a detailed schema of the fields that make up the predicate. This schema MUST define unambiguously a unique `predicateType` following the rules defined by the attestation framework. The schema must also clearly define the types of every field, with ranges for values where applicable. Further, the specification MUST contain rules for parsing the fields, as well as make recommendations for handling edge cases, where parsing or handling of fields would deviate from the standard link:https://github.com/in-toto/attestation/tree/main/spec/v1.0-draft#parsing-rules[parsing rules]. For example, it may be preferable in some cases to silently allow extra, undefined fields, while in other cases such metadata should be rejected.
+The specification MUST define unambiguously a unique `predicateType`, a shorter
+`predicateName` following the rules defined by the attestation framework, and
+contain a detailed schema of the fields that make up the predicate. The name of
```
marcelamelara, line 50: I wonder if we can be a little more precise and explicit here, still. As in, currently the "front matter" of a predicate spec contains a line for Type URI, which ends up mapping to the `predicateType`, a Version, which in some cases is/isn't included in the Type URI, and now I suppose we're adding a requirement for a `predicateName`, but it's not entirely clear where in the spec that should be indicated. I assumed this front matter section. 

Should we a) explicitly state that the Type URI + Version at the top should map to the `predicateType` in the Statement? (Please correct me if that's already indicated somewhere else). And b) Should we also explicitly mention which section of the spec we expect the predicate name to defined in?

marcelamelara, line 50: I wonder if we can be a little more precise and explicit here, still. As in, currently the "front matter" of a predicate spec contains a line for Type URI, which ends up mapping to the predicateType, a Version, which in some cases is/isn't included in the Type URI, and now I suppose we're adding a requirement for a predicateName, but it's not entirely clear where in the spec that should be indicated. I assumed this front matter section.

Should we a) explicitly state that the Type URI + Version at the top should map to the predicateType in the Statement? (Please correct me if that's already indicated somewhere else). And b) Should we also explicitly mention which section of the spec we expect the predicate name to defined in?

marcelamelara (commented):
Thanks for the updates! I have one more suggestion for clarification.

ITE/9/README.adoc
adityasaky: I took a stab at clarifying this.

ITE/9/README.adoc
```diff
@@ -45,7 +45,29 @@ The link:https://github.com/in-toto/attestation/[in-toto Attestation Framework]
...
 
-The specification MUST contain a detailed schema of the fields that make up the predicate. This schema MUST define unambiguously a unique `predicateType` following the rules defined by the attestation framework. The schema must also clearly define the types of every field, with ranges for values where applicable. Further, the specification MUST contain rules for parsing the fields, as well as make recommendations for handling edge cases, where parsing or handling of fields would deviate from the standard link:https://github.com/in-toto/attestation/tree/main/spec/v1.0-draft#parsing-rules[parsing rules]. For example, it may be preferable in some cases to silently allow extra, undefined fields, while in other cases such metadata should be rejected.
+The specification's front matter MUST unambiguously declare a predicate type, a
+version, and a shorter predicate name following the rules defined by the
+attestation framework. The predicate type along with the version maps to the
+`predicateType` field in the `Statement` layer of an in-toto attestation. The
```
marcelamelara, line 51: Minor nit: start a new paragraph after "in-toto attestation."

marcelamelara (approved):
Thanks for the updates! LGTM. I just have one very minor editorial nit at this point.

ITE/9/README.adoc
adityasaky: I split this and the example up into a separate paragraph.

ITE/9/README.adoc
```diff
@@ -45,7 +45,32 @@ The link:https://github.com/in-toto/attestation/[in-toto Attestation Framework]
...
+attestation framework. The predicate type along with the version maps to the
+`predicateType` field in the `Statement` layer of an in-toto attestation.
+
+The short predicate name is used as the name of any files related to the
+predicate specification (specification document, directories, protobuf
+definition). The name must not include version information and is expected to
+only be used as an unauthenticated hint in other contexts such as a media type
+to indicate the predicate stored in an attestation.
+
+For example, the in-toto link predicate has the type
```
trishankatdatadog, line 59:
```suggestion
For example, the in-toto link predicate has the predicateType
```

ITE/9/README.adoc
```diff
@@ -45,7 +45,32 @@ The link:https://github.com/in-toto/attestation/[in-toto Attestation Framework]
...
+`https://in-toto.io/attestation/link/vVERSION`.  Its short name is just
```
trishankatdatadog, line 60:
```suggestion
`https://in-toto.io/attestation/link/VERSION`.  Its short name is just
```

ITE/9/README.adoc
```diff
@@ -45,7 +45,32 @@ The link:https://github.com/in-toto/attestation/[in-toto Attestation Framework]
...
+`link`.  Consequently, the specification file must be named `link.md` while the
+protobuf definition must be named `link.proto`. A media type for an attestation
+file that contains a link predicate may also use `link` to indicate the
+predicate.
```
trishankatdatadog, line 64:
```suggestion
predicate (e.g., `application/vnd.in-toto.link+dsse`).
```

trishankatdatadog (commented):
Muuuuuuuuuuch better.

trishankatdatadog (approved):
LGTM but I defer to @marcelamelara and @TomHennen for final approval

marcelamelara (approved):

TomHennen (approved):
This is great, thanks!

ITE/9/README.adoc
```diff
@@ -45,7 +45,32 @@ The link:https://github.com/in-toto/attestation/[in-toto Attestation Framework]
...
 
-The specification MUST contain a detailed schema of the fields that make up the predicate. This schema MUST define unambiguously a unique `predicateType` following the rules defined by the attestation framework. The schema must also clearly define the types of every field, with ranges for values where applicable. Further, the specification MUST contain rules for parsing the fields, as well as make recommendations for handling edge cases, where parsing or handling of fields would deviate from the standard link:https://github.com/in-toto/attestation/tree/main/spec/v1.0-draft#parsing-rules[parsing rules]. For example, it may be preferable in some cases to silently allow extra, undefined fields, while in other cases such metadata should be rejected.
+The specification's front matter MUST unambiguously declare a predicate type, a
+version, and a shorter predicate name following the rules defined by the
```
JustinCappos, line 49: Do these need to be unique?  Is there a registry of these where all the used names are listed?

adityasaky: Yes, they need to be unique, good catch. We don't _define_ a formal registry but note that ITE-9 is for vetting predicates and listing them here: https://github.com/in-toto/attestation/tree/main/spec/predicates.

So, the predicate name has to be unique to not collide with an existing file. I'll make that explicit in ITE-9. The predicate type is already required to be unique in the attestation spec but I'll also make it explicit here.

adityasaky: Done!

ITE/9/README.adoc
```diff
@@ -45,7 +45,33 @@ The link:https://github.com/in-toto/attestation/[in-toto Attestation Framework]
...
+to indicate the predicate stored in an attestation.
+
+For example, the in-toto link predicate has the predicate type
+`https://in-toto.io/attestation/link/VERSION`.  Its short name is just
+`link`.  Consequently, the specification file must be named `link.md` while the
+protobuf definition must be named `link.proto`. A media type for an attestation
+file that contains a link predicate may also use `link` to indicate the
+predicate, for example `application/vnd.in-toto.link+dsse`.
+
+Further, the specification MUST include a detailed schema of the fields that
```
JustinCappos, line 67: Are they required to include an implementation of the specification parsing / handling code as well?

JustinCappos (commented):
Should we also

ITE/9/README.adoc
adityasaky: Typically the author of the predicate spec is expected to provide a protobuf definition for the predicate, which in turn provides APIs for predicates thanks to code gen.

note that this isn't a change introduced in this PR, it's only in the diff because of line wrapping / new paragraph.

adityasaky: The ITE may also not be the best place to specify other requirements for a new predicate's spec, as they may change over time. I suggest we scope the ITE to the high level workflow (how to submit a new predicate, who reviews it) and the predicate specification's contents. The attestation repo's [docs](https://github.com/in-toto/attestation/blob/main/docs/new_predicate_guidelines.md) can handle other requirements.

JustinCappos (approved):

PR #56, opened 2024-07-10
Update ITE-10 to reflect actual schema

nmiyake:
Updates schema values in ITE-10 to reflect the actual values as specified in the in-toto spec and updates the newly proposed field to match the existing style.

### DIFF
--- a/ITE/10/README.adoc
+++ b/ITE/10/README.adoc
@@ -99,15 +99,16 @@ checkout step.
 
 === Setting Predicate Type Expectations for Steps and Inspections
 
-Currently, step declarations have the following schema.
+Currently (in-toto v1.0.0), step declarations have the following
+link:https://github.com/in-toto/specification/blob/v1.0/in-toto-spec.md#431-steps[schema]:
 
 ```
 name: string
 threshold: int
-expectedMaterials: list of artifact rules
-expectedProducts: list of artifact rules
+expected_materials: list of artifact rules
+expected_products: list of artifact rules
 pubkeys: list of authorized key IDs
-expectedCommand: list of strings
+expected_command: list of strings
 ```
 
 This ITE proposes updating this schema to incorporate predicate types.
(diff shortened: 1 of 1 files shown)


ITE/10/README.adoc
```diff
@@ -99,25 +99,26 @@ checkout step.
...
 
 ```
 name: string
 command: list of strings
-expectedMaterials: list of artifact rules
-expectedProducts: list of artifact rules
-expectedPredicates: list of expectedStepPredicates
+expected_materials: list of artifact rules
+expected_products: list of artifact rules
+expected_predicates: list of expectedStepPredicates
```
nmiyake, line 121: Note that the new `expectedStepPredicate` schema below proposes `predicateType` as a field -- since this object is new this is not a backcompat break, but stylistically it would be more consistent to declare this as `predicate_type`.

nmiyake:
@adityasaky this is per discussion on https://github.com/in-toto/ITE/pull/50/files#r1651465773.

I personally prefer `camelCase` as a style and think that the core in-toto layout file format itself could stand to have a revamp of its definition (for example, `_type` field seems unneeded, and it would be nice to have a full JSON schema definition for the layout file in a single format/place rather than having it spread through the spec), but I think that changing the file format itself would be a separate independent ITE.

In the meantime, given that this ITE is based on the v1.0.0 spec, I think it's confusing for it to refer to the "existing schema" with field name formatting that doesn't match the actual spec.

ITE/10/README.adoc
adityasaky: I don't have a strong opinion stylistically here as the changes are going to be backwards incompatible anyway. @nmiyake, are you proposing we snake case this for consistency with what we already have in v1 layouts?

At the same time, I've increasingly wondered whether multiple predicates per step is actually an anti-pattern that can be specially handled for something like witness's attestation collection type (cc @jkjell). If we decide to have one predicate type per step, we may actually arrive at a case where ITE-10's changes are backwards compatible, with only the addition of new fields (predicate type), so that may influence this decision.

adityasaky: Tracking in https://github.com/in-toto/ITE/issues/58

I think we can merge this in the meantime and decide on the fields + schema / naming separately.

adityasaky:
@nmiyake sorry for the lengthy turnaround on this, could you rebase with DCO? Thanks!

trishankatdatadog (approved):

adityasaky:
Thanks @nmiyake!

adityasaky (approved):

PR #60, opened 2025-03-25
ITE-12: CDDL, CBOR, and COSE signatures

deeglaze:
I'd like for us to attempt to tackle the conciseness problem with an internet standard wire format with associated data definition language.

With CDDL defining the structure of all the predicates precisely, we can see about adding in-toto as a Tag type in the IETF CoRIM specification.

### DIFF
--- a/ITE/12/README.adoc
+++ b/ITE/12/README.adoc
@@ -1,0 +1,155 @@
+= ITE-12: CDDL, CBOR, and COSE signatures
+:source-highlighter: pygments
+:toc: preamble
+:toclevels: 2
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+.Metadata
+[cols="2"]
+|===
+| ITE
+| 0012
+
+| Title
+| CDDL, CBOR, and COSE signatures
+
+| Sponsor
+| link:https://github.com/deeglaze [Dionna Glaze]
+
+| Status
+| Active :smile:
+
+| Type
+| Standards
+
+| Created
+| 2025-02-14
+
+|===
+
+
+[[abstract]]
+== Abstract
+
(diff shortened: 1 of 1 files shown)


adityasaky:
Hi @deeglaze, sorry for the delay!

We've had earlier attempts at not being very specific about serialization formats and signing envelopes. I wonder if a subset of what you propose here can be applied to the in-toto/attestation spec so that spec is not JSON-specific. We already have ITE-5 that'll allow for using COSE.

At that point, would it require some operational pieces in in-toto/attestation to enable CDDL (i.e., if the spec is updated to be less dependent on JSON alone)?

cc @in-toto/attestation-maintainers

SantiagoTorres:
My b, I had an action item on this and never got to it. I claim that writing an informational ITE should be doable, so that we have an "encoding" of how the COSE wrapper should be implemented. In theory the ITE-5 carveout lets implementers use COSE and be compliant/follow the spec, but we could add more clarification on e.g., what're authenticated headers on in-toto if you're using a COSE wrapper...

trishankatdatadog:
Agreed with an information ITE along the lines of the successful idea of [Uptane-style protocols, operations, usage, and formats (POUFs)](https://github.com/uptane/POUFs).

deeglaze:
The downside of the "we allow other serialization formats" but the project's own rejection of anything other than targets of protoc, which is entirely Google-controlled, makes ITE-5 weak without this proposal. If there were openness to contributions to a different serialization format that aren't supported by protoc, then I wouldn't mind. My intention was to get very specific data models specified even if they're still treated as an information model for different serialization in order to have unambiguous interpretation. I don't want to take an open source project, make a bespoke format for internal efficiency, and then not be able to contribute back for what we use to be more broadly understood in the same ecosystem.

JustinCappos:
Thanks for reaching out.  I believe this is supported by the spec, but we'd
like to make this more explicit.  Right now, we were planning to make CBOR
support clear first in the DSSE spec and then look to clarify things like
this here.  So, stay tuned!

PR #61, opened 2025-08-01
ITE-13: Git Repository Attestation Storage Specification

wlynch:
Proposes an attestation storage format for how to store attestations within a Git repo.

### DIFF
--- a/ITE/13/README.adoc
+++ b/ITE/13/README.adoc
@@ -154,7 +154,7 @@ With the focus on SLSA Source provenance, there is a desire to store attestation
 
 Currently we are aware of 3 different tools that store attestations in Git repositories in slightly different ways:
 - [Sigstore Gitsign](https://github.com/sigstore/gitsign) (refs/attestations)
-- [Gittuf](https://github.com/sigstore/gitsign (refs/gittuf/attestations)
+- [gittuf](https://github.com/sigstore/gitsign (refs/gittuf/attestations)
 - [SLSA Source PoC](https://github.com/slsa-framework/slsa-source-poc) (refs/notes)
 
 Ideally, these tools should store attestations in a similar way to allow interoperability and ease of use.
--- a/ITE/13/README.adoc
+++ b/ITE/13/README.adoc
@@ -1,0 +1,194 @@
+= ITE-13: Git Repository Attestation Storage Specification
+:source-highlighter: pygments
+:toc: preamble
+:toclevels: 2
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:important-caption: :heavy_exclamation_mark:
+:caution-caption: :fire:
+:warning-caption: :warning:
+endif::[]
+
+.Metadata
+[cols="2"]
+|===
+| ITE
+| 0013
+
+| Title
+| Git Repository Attestation Storage Specification
+
+| Sponsor
+| link:https://github.com/wlynch[Billy Lynch]
+
+| Status
+| Active :smile:
+
+| Type
+| Standards
+
+| Created
+| 2025-08-01
+
+|===
+
+
+[[abstract]]
+== Abstract
+
+This specification defines how to create and store attestations for Git source code entities using the in-toto attestation framework and store them in git repositories using a standardized reference namespace.
+
+[[specification]]
+== Specification
+
+### Storage Format
+
+#### Reference Namespace
+
+Attestations MUST be stored in the `refs/attestations/<type>` namespace, corresponding to the Git object type:
+
+- **Commit attestations**: `refs/attestations/commits`
+- **Tree attestations**: `refs/attestations/trees`
+- **Blob attestations**: `refs/attestations/blobs`
+
+Additional types may be created under the `refs/attestations` namespace to store attestations about other types, even non-Git objects.
+
+#### File layout
+
+Each reference in the attestation namespace points to a Git commit containing attestation bundle files.
+The directory structure within each commit organizes attestations by the Git object hash being attested to,
+with each object hash forming a directory name containing related attestation files.
+
+```tree
+refs/attestations/commits
+├── a1b2c3d4e5f6789012345678901234567890abcd/
+│   └── attestations.intoto.jsonl
+└── b2c3d4e5f6789012345678901234567890abcdef1/
+    ├── pull_request.intoto.jsonl
+    ├── push.intoto.jsonl
+    └── deploy.intoto.jsonl
+```
+
+```tree
+refs/attestations/trees
+├── 789abc0123456789012345678901234567890def/
+│   └── attestations.intoto.jsonl
+└── def0123456789012345678901234567890abc789/
+    └── security_audit.intoto.jsonl
+```
+
+All source attestations MUST use the [Attestation bundle](./bundle.md) format.
+
+Attestations in the bundle MAY have different subjects, but they SHOULD
+all be relevant to the storage folders git object hash.
+
+#### File Naming Convention
+
+Attestation bundles SHOULD be named to create logical groupings of attestations.
+
+These may correspond to particular events:
+
+- Pre-submit (e.g. Pull Request) attestations
+- Post-submit attestations
+- CI/CD Deploy
+
+Clients should choose filenames to make it easy for other verifiers to discover information they are looking for.
+
+##### Pushing / pulling attestations
+
+Attestations are stored in [git references](https://git-scm.com/book/en/v2/Git-Internals-Git-References),
+and can be pushed or pulled like any other Git ref.
+
+The latest commit in each attestation ref represents the current state of attestations for that Git object type.
+It is the responsibility of clients to push and pull attestation references to keep them synchronized across repositories.
+
+##### Resolving conflicts
+
+Clients SHOULD be careful to avoid conflicts when editing any `refs/attestations/` ref, though normal git merge semantics apply.
+Tips for avoiding conflicts:
+
+- Choose attestation filenames that are unlikely to conflict with other automation.
+- Check for updates frequently.
+- Avoid force-pushes to avoid losing content.
+
+### Discovery Workflow
+
+#### Basic Discovery Process
+
+To discover attestations for a Git object:
+
+1. **Identify the target object**: Determine the Git object hash (commit, tree, or blob) you want to find attestations for
+2. **Check the appropriate namespace**: Look in the corresponding `refs/attestations/<type>` namespace
+3. **Navigate to the object directory**: Find the directory named with the target object's hash
+4. **Enumerate attestation files**: List all `.intoto.jsonl` files in that directory
+5. **Parse and verify**: Read the attestation bundles and verify signatures as needed
+
+#### Finding Attestations for a Git Tag
+
+To find attestations for a specific git tag, first resolve the tag to its corresponding commit SHA, then look up attestations in the commits namespace:
+
+```bash
+# Get the commit SHA for a tag
+$ git rev-parse v1.2.3
+a1b2c3d4e5f6789012345678901234567890abcd
+
+# Check for attestations about that commit
+$ git ls-tree refs/attestations/commits a1b2c3d4e5f6789012345678901234567890abcd/
+100644 blob 1234567890abcdef... attestations.intoto.jsonl
+100644 blob abcdef1234567890... release.intoto.jsonl
+```
+
+[[motivation]]
+== Motivation
+
+Source code repositories require integrity verification and provenance tracking at the git object level. This specification provides:
+
+- Git-native storage using `refs/attestations` namespace
+- Integration with existing in-toto predicate framework
+
+[[reasoning]]
+== Reasoning
+
+With the focus on SLSA Source provenance, there is a desire to store attestations about Git objects directly in the repository.
+
+Currently we are aware of 3 different tools that store attestations in Git repositories in slightly different ways:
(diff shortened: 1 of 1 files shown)


wlynch:
cc @adityasaky @puerco @TomHennen @patzielinski

ITE/13/README.adoc
```diff
@@ -0,0 +1,194 @@
...
+├── 789abc0123456789012345678901234567890def/
+│   └── attestations.intoto.jsonl
+└── def0123456789012345678901234567890abc789/
+    └── security_audit.intoto.jsonl
+```
+
+All source attestations MUST use the [Attestation bundle](./bundle.md) format.
+
+Attestations in the bundle MAY have different subjects, but they SHOULD
+all be relevant to the storage folders git object hash.
```
patzielinski: Might be a typo?
```suggestion
all be relevant to the storage folder's git object hash.
```

ITE/13/README.adoc
```diff
@@ -0,0 +1,194 @@
...
+- CI/CD Deploy
+
+Clients should choose filenames to make it easy for other verifiers to discover information they are looking for.
+
+##### Pushing / pulling attestations
+
+Attestations are stored in [git references](https://git-scm.com/book/en/v2/Git-Internals-Git-References),
+and can be pushed or pulled like any other Git ref.
+
+The latest commit in each attestation ref represents the current state of attestations for that Git object type.
```
patzielinski, line 103: Do we make any statement that the storage of attestations must strictly grow (i.e. each new commit has all previous attestations inside plus any new ones)?

ITE/13/README.adoc
```diff
@@ -0,0 +1,194 @@
...
+[cols="2"]
+|===
+| ITE
+| 0013
+
+| Title
+| Git Repository Attestation Storage Specification
+
+| Sponsor
+| link:https://github.com/wlynch[Billy Lynch]
```
SantiagoTorres, line 23: I think the sponsor here should be an ITSC editor, rather than the author

ITE/13/README.adoc
```diff
@@ -0,0 +1,194 @@
...
+
+#### Basic Discovery Process
+
+To discover attestations for a Git object:
+
+1. **Identify the target object**: Determine the Git object hash (commit, tree, or blob) you want to find attestations for
+2. **Check the appropriate namespace**: Look in the corresponding `refs/attestations/<type>` namespace
+3. **Navigate to the object directory**: Find the directory named with the target object's hash
+4. **Enumerate attestation files**: List all `.intoto.jsonl` files in that directory
+5. **Parse and verify**: Read the attestation bundles and verify signatures as needed
```
SantiagoTorres, line 125: I am trying to picture this in my head. This is to say that an attestation collection is inside a worktree? is this separate from the standard worktree? or is storage tangential and we don't care?

ITE/13/README.adoc
```diff
@@ -0,0 +1,194 @@
...
+If your implementation includes code, include this section describing how your
+changes will be tested.
+
+[[prototype-implementation]]
+== Prototype Implementation
+
+If your ITE includes code changes, provide some links to the prototype
+implementation of these changes.
+
+TODO (we will use one or all of the existing tools as a prototype implementation)
```
SantiagoTorres, line 190: I wonder if instead of TODO we could just point out to all of these frameworks that are, admittedly, quite mature already.

SantiagoTorres (commented):
I think this is cool! I'd like to move forward a bit. I think some clarification and we can start gathering comments

ITE/13/README.adoc
adityasaky: Or both?

ITE/13/README.adoc
```diff
@@ -0,0 +1,194 @@
...
+
+#### Reference Namespace
+
+Attestations MUST be stored in the `refs/attestations/<type>` namespace, corresponding to the Git object type:
+
+- **Commit attestations**: `refs/attestations/commits`
+- **Tree attestations**: `refs/attestations/trees`
+- **Blob attestations**: `refs/attestations/blobs`
+
+Additional types may be created under the `refs/attestations` namespace to store attestations about other types, even non-Git objects.
```
adityasaky: Might be worth thinking about the scope here: if the attestations are about the git objects in the same repo, it makes sense for them to live in `refs/attestations/<>`, i.e., out of the way of "regular" repo contents in `main` etc. If we want to store attestations for non Git objects, does it need to go in the special ref namespace in the first place?

ITE/13/README.adoc
```diff
@@ -0,0 +1,194 @@
...
+Each reference in the attestation namespace points to a Git commit containing attestation bundle files.
+The directory structure within each commit organizes attestations by the Git object hash being attested to,
+with each object hash forming a directory name containing related attestation files.
+
+```tree
+refs/attestations/commits
+├── a1b2c3d4e5f6789012345678901234567890abcd/
+│   └── attestations.intoto.jsonl
+└── b2c3d4e5f6789012345678901234567890abcdef1/
+    ├── pull_request.intoto.jsonl
```
adityasaky, line 68: Should we allow multiple blobs? Can we consistently have `attestations.intoto.jsonl` that is a union of all the jsonl files in this example? That might simplify client logic. We can determine later if multiple files are necessary, perhaps.

We could even argue for `commitID.intoto.jsonl` instead of having a directory for the commit ID with a single item in it, but that might be problematic to extend.

ITE/13/README.adoc
```diff
@@ -0,0 +1,194 @@
...
+
+```tree
+refs/attestations/trees
+├── 789abc0123456789012345678901234567890def/
+│   └── attestations.intoto.jsonl
+└── def0123456789012345678901234567890abc789/
+    └── security_audit.intoto.jsonl
+```
+
+All source attestations MUST use the [Attestation bundle](./bundle.md) format.
```
adityasaky: link is local rather than to the attestations repo.

ITE/13/README.adoc
```diff
@@ -0,0 +1,194 @@
...
+
+Attestations in the bundle MAY have different subjects, but they SHOULD
```
adityasaky, line 83: I suggest bumping this up to MUST to keep things simpler. Attestations in the bundle may have 1+ subjects but _at least_ one of those subjects must be for the object whose directory it's in to avoid scenarios where an attestation is assumed to apply to the Git object whose directory it's in.

ITE/13/README.adoc
wlynch: Up to the user? This is intentionally written to not be prescriptive, the same way you can extend `refs/...` in generally to store information about what ever you want.

ITE/13/README.adoc
wlynch: I think we should. While it might make verification more complex, it's much easier as an implementor to stake claim of a particular file namespace without care of other integrations instead of having to do read-modify-write workflows.

e.g. the gittuf app might always write observations as `github_<app-id>.json`.

ITE/13/README.adoc
wlynch: Fixed.

ITE/13/README.adoc
wlynch: From a structural standpoint it doesn't really matter if these don't match?
Verifiers should validate but the biggest issue if you don't follow this is discoverability is hindered.

We want to avoid cruft here, but the security properties still holds if this is mucked with, so SHOULD feels appropriate?

ITE/13/README.adoc
wlynch: Done.

ITE/13/README.adoc
wlynch: No - I kinda few this like OCI attestations in that that probably makes the most sense, but there's no harm in removing them beyond you may not be able to verify anymore (which sometimes may be a feature).

ITE/13/README.adoc
wlynch: I'm not sure I fully understand, but the ref structure is more about discoverability of attestations - e.g. the same way for OCI you can use referrers / `<digest>.att`, you should be able to use `refs/attestations/...` to locate attestations in git repos. From there is a matter of iterating through the candidates to find what you need - either an attestation for a certain type/subject/signer, etc.

ITE/13/README.adoc
wlynch: Part of the motivation for this was to settle on a common way that these tools can interop with each other by storing things in similar formats. Some use refs/attestations (but different format) today, some use git notes, etc.

ITE/13/README.adoc
```diff
@@ -0,0 +1,194 @@
...
+- Integration with existing in-toto predicate framework
+
+[[reasoning]]
+== Reasoning
+
+With the focus on SLSA Source provenance, there is a desire to store attestations about Git objects directly in the repository.
+
+Currently we are aware of 3 different tools that store attestations in Git repositories in slightly different ways:
+- [Sigstore Gitsign](https://github.com/sigstore/gitsign) (refs/attestations)
+- [Gittuf](https://github.com/sigstore/gitsign (refs/gittuf/attestations)
```
patzielinski:
```suggestion
- [gittuf](https://github.com/sigstore/gitsign (refs/gittuf/attestations)
```

patzielinski (approved):
No further suggestions from me from the gittuf perspective :slightly_smiling_face:.

ITE/13/README.adoc
```diff
@@ -0,0 +1,194 @@
...
+| 0013
+
+| Title
+| Git Repository Attestation Storage Specification
+
+| Sponsor
+| link:https://github.com/wlynch[Billy Lynch]
+
+| Status
+| Active :smile:
```
adityasaky, line 26: Can we drop this to draft so we can merge and iterate?

ITE/13/README.adoc
adityasaky: @wlynch can you list me as a cosponsor?

ITE/13/README.adoc
```diff
@@ -0,0 +1,194 @@
...
+
+### Storage Format
+
+#### Reference Namespace
+
+Attestations MUST be stored in the `refs/attestations/<type>` namespace, corresponding to the Git object type:
+
+- **Commit attestations**: `refs/attestations/commits`
+- **Tree attestations**: `refs/attestations/trees`
+- **Blob attestations**: `refs/attestations/blobs`
```
adityasaky, line 53: can we have attestations about tags? I could see that being valuable (e.g., attesting to a release).

ITE/13/README.adoc
adityasaky: Ack, resolving.

ITE/13/README.adoc
adityasaky: I'm happy with SHOULD at this point. I think the sentence still needs some tweaking because you can have folders that are not git objects and thus not hashed as such?

ITE/13/README.adoc
```diff
@@ -0,0 +1,194 @@
...
+#### File Naming Convention
+
+Attestation bundles SHOULD be named to create logical groupings of attestations.
+
+These may correspond to particular events:
+
+- Pre-submit (e.g. Pull Request) attestations
+- Post-submit attestations
+- CI/CD Deploy
+
```
adityasaky, line 95:
```suggestion
These may correspond to particular events, such as but not limited to:

- Pre-submit (e.g. Pull Request) attestations
- Post-submit attestations
- CI/CD Deploy
```

ITE/13/README.adoc
adityasaky: In case of a tag, I'd use the ID rather than the name.

ITE/13/README.adoc
adityasaky: My read is that attestations are in a worktree tracked by `refs/attestations/<whatever>` and this is a distinct worktree from the one(s) tracked by `refs/heads/main` and other branches. @SantiagoTorres does this clarify?

ITE/13/README.adoc
```diff
@@ -0,0 +1,194 @@
...
+1. **Identify the target object**: Determine the Git object hash (commit, tree, or blob) you want to find attestations for
+2. **Check the appropriate namespace**: Look in the corresponding `refs/attestations/<type>` namespace
+3. **Navigate to the object directory**: Find the directory named with the target object's hash
+4. **Enumerate attestation files**: List all `.intoto.jsonl` files in that directory
+5. **Parse and verify**: Read the attestation bundles and verify signatures as needed
+
+#### Finding Attestations for a Git Tag
+
+To find attestations for a specific git tag, first resolve the tag to its corresponding commit SHA, then look up attestations in the commits namespace:
```
adityasaky, line 129: Interesting. I advocate for having separate tag attestations. A user may choose to resolve the tag's target object and load the attestations corresponding to that, but I think there's value in giving a tag object its own space for attestations that cannot be confused with those of the underlying object. For example, consider an attestation tracking approval at the commit level. You may want a separate approval attestation to create a tag over that commit and putting that in a separate place reduces a potential footgun IMO.

ITE/13/README.adoc
adityasaky: I think we could quite easily update gittuf to store attestations in the format proposed here (in addition to gittuf's `refs/gittuf/attestations` which is a load bearing ref at this point).

adityasaky (commented):
Sorry for the delay in coming back to this. I'd like to get this merged soon as a draft so we can proceed with writing proofs of concepts. @wlynch let me know if it'd help to work on my comments together to hopefully get this to a merge-and-iterate state soonish. :)

PR #62, opened 2025-08-20
Update ITE-10: change step predicates and add selectors

marcelamelara:
This PR makes revisions to the current text of ITE-10 based on practical experiences and case studies verifying in-toto attestations:

* 1-1 predicate type-to-step mapping: This simplifies step definitions in layouts as well as attestation verifier implementations.
* Materials and products field selectors: Since the names of the fields for step materials and products are no longer consistent across predicate types, these selectors allow layout authors to indicate to verifiers where to extract this information from within attestations.

These changes were first proposed at [OpenSSF Community Day NA '25](https://sched.co/1zhne).

### DIFF
--- a/ITE/10/README.adoc
+++ b/ITE/10/README.adoc
@@ -123,14 +123,20 @@ expected_predicate: expectedStepPredicate
 
 The `expected_inputs` and `expected_outputs` fields replace the original
 `expected_materials` and `expected_products` fields, respectively, to generalize
-theri semantics. This change maintains the core model of the original in-toto
-layout associating a single predicate with a single step (rather than a single
-command). It is also still true that if two attestations do not share the input
-and output artifacts, they must belong to different steps. The `expected_inputs`
-field retains the current semantics that all steps have inputs regardless of
-predicate type, so it applies to all predicates, informational and
-transformational, but does the name of the field containing the inputs must
-still be indicated within the `expected_predicate` information.
+their semantics. This change maintains the core model of the in-toto v1 layout
+associating a single predicate with a single step (rather than a single
+command). It is also still true that, `expected_inputs` and `expected_outputs`
+specify a list of artifact rules against which step inputs and outputs must be
+checked. Thus, if two attestations do not share the input and output artifacts,
+it is still true that they must belong to different steps.
+
+The `expected_inputs` field retains the current semantics that all steps have
+inputs regardless of predicate type, so it applies to all predicates,
+informational and transformational, but does the name of the field containing
+the inputs must still be indicated within the `expected_predicate` information.
+Of course, as informational predicates do not specify output artifacts, the
+`expected_outputs` field does not apply to them and may be omitted or left empty
+for the corresponding step.
 
 Each `expectedStepPredicate` object has the following schema.
 
--- a/ITE/10/README.adoc
+++ b/ITE/10/README.adoc
@@ -118,35 +118,37 @@ name: string
 command: list of strings
 expected_materials: list of artifact rules
 expected_products: list of artifact rules
-expected_predicates: list of expectedStepPredicates
+expected_predicate: expectedStepPredicate
 ```
 
 Each `expectedStepPredicate` object has the following schema.
 
 ```
 predicateType: string
+materials_selector: string
+products_selector: string
 functionaries: list of authorized functionary key IDs
 threshold: int
 ```
 
 The `predicateType` field contains the matching `TypeURI` of the predicate
-expected for the step. The original `pubkeys` field is updated to
+expected for the step. To support both transformational and informational
+predicates that may use different field names to record step materials and/or
+products, the `materials_selector` and `products_selector` fields to indicate
+the specific ITE-6 attestation field names verifiers should check for materials
+and products, respectively. The original `pubkeys` field is updated to
 `functionaries` to generalize signing semantics. Finally, `threshold` applies
 per predicate type rather than the step as a whole. Note that this schema
 supports the in-toto Attestation Framework while retaining all the capabilities
 of in-toto 1.0 layouts.
 
-This change maintains the core model of the original in-toto layout while
-allowing for multiple predicate types to be associated for a single step. Note
-that while this schema allows for multiple predicate types, the expectation is
-that they share the same materials and products--if two attestations do not
-share the input and output artifacts, they must belong to different steps. Of
-course, as informational predicates do not specify artifact products, the
-`expectedProducts` field does not apply to them. In a step with exclusively
-informational predicates, the `expectedProducts` field may be omitted or left
-empty. In a step with a mixture of transformational and informational
-predicates, the field only applies to the transformational predicates. The
-`expectedMaterials` field functions as it currently does, and applies to all
+This change maintains the core model of the original in-toto layout associating
+a single predicate with a single step (rather than a single command). It is
+also still true that if two attestations do not share the input and output
+artifacts, they must belong to different steps. Of course, as informational
+predicates do not specify artifact products, the `expectedProducts` field does
+not apply to them and may be omitted or left empty for the corresponding step.
+The `expectedMaterials` field functions as it currently does, and applies to all
 predicates, informational and transformational.
 
 === Supporting Sublayouts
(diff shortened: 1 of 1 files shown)


ITE/10/README.adoc
```diff
@@ -118,35 +118,37 @@ name: string
...
 expected_products: list of artifact rules
-expected_predicates: list of expectedStepPredicates
+expected_predicate: expectedStepPredicate
 ```
 
 Each `expectedStepPredicate` object has the following schema.
 
 ```
 predicateType: string
+materials_selector: string
```
trishankkarthik: Could we call this `inputs_selector` instead, please?

ITE/10/README.adoc
```diff
@@ -118,35 +118,37 @@ name: string
...
+products_selector: string
```
trishankkarthik: Could we call this `outputs_selector` instead, please?

ITE/10/README.adoc
```diff
@@ -118,35 +118,37 @@ name: string
...
 functionaries: list of authorized functionary key IDs
 threshold: int
 ```
 
 The `predicateType` field contains the matching `TypeURI` of the predicate
-expected for the step. The original `pubkeys` field is updated to
+expected for the step. To support both transformational and informational
+predicates that may use different field names to record step materials and/or
+products, the `materials_selector` and `products_selector` fields to indicate
+the specific ITE-6 attestation field names verifiers should check for materials
```
trishankkarthik: I think we want to expand just a bit on how verifiers should check materials/inputs and products/outputs: specifically while checking [v1-style artefact rules](https://github.com/in-toto/specification/blob/v1.0/in-toto-spec.md#433-artifact-rules).

ITE/10/README.adoc
```diff
@@ -118,35 +118,37 @@ name: string
...
-empty. In a step with a mixture of transformational and informational
-predicates, the field only applies to the transformational predicates. The
-`expectedMaterials` field functions as it currently does, and applies to all
+This change maintains the core model of the original in-toto layout associating
+a single predicate with a single step (rather than a single command). It is
+also still true that if two attestations do not share the input and output
+artifacts, they must belong to different steps. Of course, as informational
+predicates do not specify artifact products, the `expectedProducts` field does
+not apply to them and may be omitted or left empty for the corresponding step.
+The `expectedMaterials` field functions as it currently does, and applies to all
```
trishankkarthik: Sorry, not clear to me: do you mean that `expectedMaterials` now refer to subjects by default unless specified otherwise?

trishankkarthik (commented):
Thanks very much for these simple but highly effective improvements, Marcela! Just a few comments...

ITE/10/README.adoc
marcelamelara: I also prefer `inputs_selector` but given that we still use `materials` and `products` for the artifact rules, I think the selector names should match. I wouldn't be opposed to switching to `inputs` and `outputs` overall, but wasn't sure if that'd be too big a change.

ITE/10/README.adoc
marcelamelara: This is text from @adityasaky . The intent here is that all steps have `materials`, regardless of the predicate type. It says nothing about which fields those are.

marcelamelara: We can add a clarification about this.

ITE/10/README.adoc
```diff
@@ -158,24 +160,29 @@ verification workflow by presenting the sublayout file in place of the link
...
 artifact rules are updated to support
 link:https://github.com/in-toto/attestation/blob/main/spec/v1/resource_descriptor.md[ResourceDescriptor]
 defined in the in-toto Attestation Framework. The `name` field in a
-ResourceDescriptor object is used for pattern matching and the `digests` field
+ResourceDescriptor object is used for pattern matching and the `digest` field
 is used to verify equality of two artifacts.
 
+As in in-toto v.10, the `layoutEnvelope` and `layoutKeys` contain the signed
+layout and the signer's keys. The generalized `attestations` parameter is the
+set of in-toto attestations (with any predicate) to be checked against the
```
marcelamelara, line 191: @trishankkarthik I'm assuming the `attestations` being input into the Verification flow is the bundle put together by the bundle compiler? And we can also assume there's one attestation in this bundle per step? Want to make sure we're making the same assumptions here.

ITE/10/README.adoc
trishankkarthik: I think since we already changed names for some big things (like attestations instead of links and so on), we might as well take this opportunity to rename this especially because it makes things so much clearer. WDYT?

ITE/10/README.adoc
trishankkarthik: Yes, please, thanks!

ITE/10/README.adoc
trishankkarthik: Yes :) Although I don't think we need to talk about the Bundle Compiler here at all, just the Attestation Bundle. The one Attestation (which implies one Predicate Type) per Step makes sense to me.

marcelamelara: Ok cool, yeah, I agree. I wasn't going to mention the bundle compiler at this point, just wanted to have the right mental model for when we do talk about it :)

trishankatdatadog (approved):
💯

marcelamelara:
@adityasaky PTAL

ITE/10/README.adoc
```diff
@@ -116,38 +116,48 @@ This ITE proposes updating this schema to incorporate predicate types.
...
+Of course, as informational predicates do not specify output artifacts, the
+`expected_outputs` field does not apply to them and may be omitted or left empty
+for the corresponding step.
+
 Each `expectedStepPredicate` object has the following schema.
 
 ```
-predicateType: string
+predicate_type: string
+inputs_selector: string
```
adityasaky, line 145: :eyes: this is interesting. Do we need outputs_selector?

Two cases:
informational -> inputs is subject, no outputs
transformational -> inputs is predicate specific, outputs is subject

we could have just the inputs_selector, with options `subject` and `predicate.<field>` where the field is set. If it's `predicate.<field>` we can safely assume the contents of subject make up the output. If input_selector is `subject`, then there's no output.

ITE/10/README.adoc
```diff
@@ -187,30 +203,27 @@ func Verify(layoutEnvelope, layoutKeys, attestations, now) {
...
-            } else {
-                verifiedClaims := verifyAttestationSignatures(predicate.Functionaries, predicateAttestations)
-                if len(verifiedClaims) < threshold {
-                    panic
-                }
-                for _, attestation := range verifiedClaims {
-                    applyMaterialRules(step.ExpectedMaterials, attestation.Materials, attestations)
-                    applyProductRules(step.ExpectedProducts, attestation.Products, attestations)
-                }
+        predicate := step.ExpectedPredicate {
```
adityasaky:
```suggestion
        predicate := step.ExpectedPredicate
```

ITE/10/README.adoc
```diff
@@ -158,24 +168,30 @@ verification workflow by presenting the sublayout file in place of the link
...
+discussed further on in <<Custom Verifiers>>.
 
 === Verifying Artifact Rules
 
-The verification workflow in the in-toto specification matches links to use by
-the name of the step and the key ID used to sign it. This part of the workflow
-is updated to support matching on predicate types as well. In addition,
+The verification workflow in the in-toto v1.0 specification matches links to
+use by the name of the step and the key ID used to sign it. This part of the
+workflow is updated to support matching on predicate types as well. In addition,
```
adityasaky, line 182: side note: we don't specify how to match on predicate types; we should consider this but separately.

ITE/10/README.adoc
```diff
@@ -158,24 +168,30 @@ verification workflow by presenting the sublayout file in place of the link
...
 artifact rules are updated to support
 link:https://github.com/in-toto/attestation/blob/main/spec/v1/resource_descriptor.md[ResourceDescriptor]
 defined in the in-toto Attestation Framework. The `name` field in a
-ResourceDescriptor object is used for pattern matching and the `digests` field
+ResourceDescriptor object is used for pattern matching and the `digest` field
 is used to verify equality of two artifacts.
 
+As in in-toto v.10, the `layoutEnvelope` and `layoutKeys` contain the signed
```
matglas: I think this should be v1.0.

ITE/10/README.adoc
trishankatdatadog: Good catch! Actually, we talk about [I/O selectors in the Policy Framework draft](https://docs.google.com/document/d/1Tr6jCtTcTB8KpDMp2I1PFuxP7A2nQa5BeuwOy05iuMw/edit?pli=1&tab=t.0#bookmark=id.4g7eqaljs29k).

ITE/10/README.adoc
trishankatdatadog:
```suggestion
As in in-toto v1.0, the `layoutEnvelope` and `layoutKeys` contain the signed
```

ITE/10/README.adoc
```diff
@@ -158,24 +168,30 @@ verification workflow by presenting the sublayout file in place of the link
...
-expected predicates in the process. Note that this expects the verifier to be
-capable of generating distinct predicates--the in-toto v1.0 verifier was only
-aware of links and sublayout verification therefore generated an in-toto summary
-link. This is discussed further on in <<Custom Verifiers>>.
+declared with an expected predicate type and when a layout is encountered, the
+verifier recursively verifies the sublayout, generating the expected predicates
+in the process. Note that this expects the verifier to be capable of generating
+distinct predicates--the in-toto v1.0 verifier was only aware of links and
+sublayout verification therefore generated an in-toto summary link. This is
+discussed further on in <<Custom Verifiers>>.
```
JustinCappos, line 176: Where is the discussion of the supported languages and custom verifiers?  I'd love to have us choose something really small and sane here which is unlikely to have any security issues.  

Usually running complex code during signature checking is a bad idea.  Since this is running potentially attacker provided code during verification, this should be really locked down.  (I acknowledge and regret how inspections work in the current in-toto layouts, so I'm mostly pointing the finger at myself here.)

trishankkarthik: That discussion is not here: this is specific to just: (1) fixing a step to exactly one predicate type, and (2) introducing I/O Selectors.

The general discussion on policies is [here](https://docs.google.com/document/d/1Tr6jCtTcTB8KpDMp2I1PFuxP7A2nQa5BeuwOy05iuMw/edit?usp=sharing). We do have not a list of supported policy languages and engines, and are not likely to do so. Please see that doc for our rationale on why we want flexibility.

JustinCappos: I took a look at the document and would love to have more discussion about the pros and cons.  Happy to do this in the document, on this issue, or in another format.  I do think there needs to be more of a comparative discussion resulting from this, regardless of the eventual design choice.

trishankkarthik: Why don't we discuss over a Zoom/Meet and go from there? I'll send out an invitation.

ITE/10/README.adoc
marcelamelara: This is an interesting suggestion. My inclination is to specify things as explicitly as possible so consumers of policies don't need to bake in implicit assumptions into their verification flow. How do others feel about removing `outputs_selector` ?

ITE/10/README.adoc
```diff
@@ -116,38 +116,48 @@ This ITE proposes updating this schema to incorporate predicate types.
...
+associating a single predicate with a single step (rather than a single
+command). It is also still true that, `expected_inputs` and `expected_outputs`
+specify a list of artifact rules against which step inputs and outputs must be
+checked. Thus, if two attestations do not share the input and output artifacts,
+it is still true that they must belong to different steps.
+
+The `expected_inputs` field retains the current semantics that all steps have
+inputs regardless of predicate type, so it applies to all predicates,
+informational and transformational, but does the name of the field containing
+the inputs must still be indicated within the `expected_predicate` information.
```
adityasaky, line 136: There's a grammatical issue here and I'm not quite sure what the fix is.

ITE/10/README.adoc
adityasaky: As things stand, I lean towards merging this in with `outputs_selector`. We might be better positioned to take a call once we PoC this in attestation-verifier perhaps.

adityasaky (approved):
I suggest we land this. My only comment with substance is a grammatical fix.

PR #63, opened 2026-08-09
ITE-0000: tiers and an evidence criterion for predicates

astrogilda:
Restructures ITE-0000 against the sixteen review comments on this pull request.

Determinate is defined before first use. The criterion is renamed and outlines its three items in place, with E1, E2 and E3 as their own sub-sub-sections. RFC 8126 frames the tiers before they are named, and a lifecycle section names every state and draws the transitions. The testing strategy is removed.

`wc -lw ITE/0000/README.adoc` reports 252 lines and 1,588 words on this branch, against 832 and 6,550 at the revision you reviewed. Every normative requirement is preserved, and the sixteen replies on this pull request each name where its requirement now lives.


ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
...
+the repository should become a hosting location for predicates, and whether the
+maintainers have the expertise for what gets submitted.
+
+This ITE proposes the tiers and the rule that sorts a submission into one. The rule
+is evidence a reviewer can check: a conformance corpus with negative controls, an
+implementation built from the specification text by someone other than its author,
+and published run results including one taken blind, released as a signed bundle.
+Three doors lead into vetted, so more than one kind of submitter can reach it.
+
+Publication is gated on none of it. A submission offering none of it is listed in
```
SantiagoTorres: This is rather terse, could you care to re-phrase? I'm not sure what it means

ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
...
+contrib, which this ITE treats as a full-status home. The evidence is required only
+to claim vetted, and it is chosen so a maintainer who is not an expert in the
+submission's domain can still check it.
+
+[[specification]]
+== Specification
+
+=== Scope
```
SantiagoTorres, line 31: This is a bit too verbose. I'm not able to fully tell what is going on. Could you please simplify? Overall on these less is more in that we can focus on being precise/succinct.

ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
...
+
+ITE-9 fixed what a predicate submission must contain. It does not define what makes
+one acceptable, and that missing half is why the meeting's two-tier idea has nothing
+to sort on. This ITE supplies it. The ITE-9 document format, namespace assignment and the
+optional protobuf definition are all left exactly as they are.
+
+The criterion is orthogonal to subject-matter scoping and composes with it. The same
+meeting recorded a scoping proposal, "Aditya: Vetted could be for fundamental SDLC
+types that lend themselves to reuse", which asks what a type is _about_; this asks
+whether a specification is _determinate_. If the project adopts a subject-matter
```
SantiagoTorres: I think the term determinate may be useful, although I'm not sure it's the correct one. From what I can gather, you're trying to ask: "is the predicate not only useful but complete/all-encompassing/properly scoped?" is that correct? I wonder if defining *determinate in the context of this document* would be useful to help readability

ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
...
+
+`contrib`:: A predicate type that is hosted and listed. Admission requires what it
+requires today: a specification in the ITE-9 format, and a maintainer check that it
+is not a duplicate and not actively harmful. In the words ITE-9 already uses of the
+library, types listed here "are not mandated or endorsed".
+
+`vetted`:: A predicate type whose specification is determinate and has been
+independently implemented. Admission requires everything contrib requires, plus the
+evidence below through any of the three doors, plus any subject-matter scope the
+project adopts.
```
SantiagoTorres: I don't have a strong opinion on this, but other projects tend to use things like "core" to refer to the first class (i.e., ones that are officially endorsed or fully adopted) and the second "contrib" (i.e., ones that are *contributed* by the broader community but may have seen less scrutiny). I wonder if leaning on terminology like this will make it more intuitive

ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
...
+
+Every submission lands in contrib by default and is listed; failing the criterion is
+not a ground for rejection. In the registry vocabulary of RFC 8126, contrib is a
+Specification Required policy and vetted is Expert Review against a published
+criterion.
```
SantiagoTorres: I think the context of RFC  8126 is probably useful and should come earlier to help us with framing. 

Further, I think that updating the promotion ladder to be a bit more explicit would be helpful. For example: what is ground for rejection? perhaps a lifecycle section can help us how predicates will be promoted/demoted/decomissioned/rejected/staged?

ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
...
+| on request, with a protobuf definition
+|===
+
+Bindings are not a tier privilege and tier does not control namespace. Nine of the
+thirteen listed types have generated Go bindings and five carry URIs in namespaces
+this project does not control, so making either a consequence of tier would create a
+rule the current directory violates on the day it ships.
+
+[[criterion]]
+=== The criterion
```
SantiagoTorres: I think this section should be called "Acceptance and Promotion Criterion" or something along those lines.

ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
...
+
+To be placed in vetted, a predicate type MUST ship three items of evidence in a
+signed bundle. Each is published, each can be checked by anyone, and none requires
```
SantiagoTorres: signed by who? with what?

ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
...
+the checker to be an expert in the predicate's domain.
```
SantiagoTorres: Can you outline the three items here ?

ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
...
+
+E1: a conformance corpus with negative controls::
```
SantiagoTorres: I wonder if E1, E2, E3 here and below would be better suited as subsubsubsubsections

ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
...
+
+Where a schema is wholly delegated to an external specification that maintains its
+own conformance suite, E1 and E2 are satisfied by citing that suite as the index
+record's normative reference, and the submission ships vectors only for the in-toto
+binding layer. Three of the thirteen listed types are that shape, and this is how the
+conformance work of SPDX, CycloneDX and any future body counts here without being
+rebuilt here.
+
+[[doors]]
+=== Three doors into vetted
```
SantiagoTorres: Please rename to: "Promotion from Contrib to Vetted" or so. I think this sounds cool but is a bit left-field as well

ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
...
+
+A single gate admits a single kind of submitter. Each door supplies the independence
+E2 is about from a different side.
```
SantiagoTorres: is it door or gate? I'd prefer we say there are three paths instead

ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
...
+Demotion and deprecation:: A vetted type whose corpus ceases to pass, or which is
+superseded, moves to `deprecated` with a pointer to its replacement. A hosted type
+found to be harmful moves to `withdrawn`. Neither path exists today and both are
+needed regardless of what this ITE decides about tiers.
+
+Contest:: Every decision under this criterion is contestable with evidence: a tier
+decision by producing the evidence the criterion names, a demotion by re-running the
+corpus green and linking the run. A contest does not require the maintainers to
+re-open the domain judgement, nor the contesting party to be an expert in anything
+except their own type.
```
SantiagoTorres: This is very useful information, but I'm tempted to frame it around "lifecycle of a predicate" perhaps with a flow diagram that shows us how things move between. As it stands, this outlines transitional phases and their processes, it'd be useful to know what states they transition to/from

ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
...
+surface: a Security Architect, a Pragmatic Developer and an Incidental Consumer. It is
+the third who will read vetted as "safe". The directory MUST therefore define vetted
+as meaning the specification is determinate and has been independently implemented,
+and state that adjacent to both listings.
+
+Evidence can be fabricated. Signing binds each run to a key rather than to a claim in
+a pull-request description, and the corpus digest prevents a corpus being rewritten
+after a run it did not pass. Neither prevents a determined author manufacturing a
+second implementer, which no process rule can.
+
```
SantiagoTorres: I wonder if one risk being introduced is that people adopt a predicate that's fundamentally broken from contrib. I'm not sure if this is something we can address but definitely something we need to.

On a second though, I'm not sure if that's what you mean with `trust signals are over-read`, but the text is a bit terse as I was saying....

ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
...
+[[infrastructure]]
+== Infrastructure Requirements
+
+Mostly existing machinery. The repository already runs CI across Go, Python, Java and
+Rust bindings and already documents a minimum testing expectation; a conformance
+corpus is a set of JSON Statements exercised by that harness and the CI addition is a
+job that runs it. New, and small: the index file, its schema and its generator, plus
+the five other artifacts named in the companion. No hosting cost change, since a
+corpus may live in the submitter's own repository pinned by digest.
```
SantiagoTorres: I take this to mean: we only need to run a CI job that creates the index upon merge, but also write this job, right? can you state this more explicitly ?

ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
...
+. Twelve months after adoption, either the submissions promoted under it are only
+  those authored by the people who proposed it, or fewer than two types authored by
+  unrelated parties have been promoted through any of the three doors.
+. The criterion marks as failing a type demonstrably interoperating in the field
+  between independent producers and consumers.
+. Any of the six artifacts named in the companion is unpublished one release cycle
+  after adoption. The claim that a submitter can self-check is false until they exist.
+
+Each condition names a date, a count or an artifact, so whether it has fired is a
+matter of record rather than opinion.
```
SantiagoTorres: I think this section is not needed, as we don't need a testing strategy for this process ITE

SantiagoTorres (changes requested):
Hi @astrogilda , thanks for the submission! I left some comments that I think we can address (or of course discuss) before we move into draft. I overall think this is a good start!

astrogilda:
Thanks Santiago, this was a genuinely useful review. An earlier version of this comment said I had pushed changes for all fifteen of your comments. Nothing had been pushed and nothing is pushed now: the branch still stands at 2edc3955 from 9 August, one file and 405 lines. What follows is what I am changing and why, so you can overrule any of it before it lands rather than after.

On the terse paragraph at the end of the abstract and the verbose Scope section, both were fair. Nothing here changes who gets published: a submission that meets today's bar is accepted and listed exactly as it is today, and the evidence is only what a type needs in addition if it wants the vetted label. Scope becomes a "this ITE changes / this ITE does not change" pair, which is shorter and easier to check against.

On "determinate", your reading was "is the predicate not only useful but complete, all-encompassing, properly scoped?", which is not what I meant, and the fact that it was the natural reading is the bug. It gets its own definition before first use: a specification is determinate when two competent readers, implementing from the text alone and without consulting each other, reach the same accept-or-reject decision on the same input. Two clarifications follow from that and your reading makes them necessary. A determinate specification can still be narrow, incomplete, or a bad idea, and a thorough, well-scoped, genuinely useful specification can still be indeterminate if two readers disagree about whether a field is required. Determinacy measures how precisely the text pins down behaviour; how much ground it covers is a separate axis. That distinction is load-bearing for the whole proposal, so thank you for catching that it was never stated.

On core versus contrib, I would rather write the tradeoff into a note and let the Editors decide it, since a rename is a mechanical substitution at this point. Two things bear on it. `docs/new_predicate_guidelines.md` already says "our list of vetted predicates", so vetted is the incumbent term and keeping it renames nothing already published, whereas core renames an existing published concept in a document that is not otherwise proposing to touch it. And on contrib, which is the right instinct and matches the convention you are pointing at, Debian's archive layout already uses contrib for packages depending on non-free ones, which carries a faintly second-class sense this ITE spends effort arguing against. The meeting's other suggestion, community, carries no such freight. I have no strong preference.

On RFC 8126 coming earlier and the promotion ladder being more explicit, agreed on both, and the second is the largest change in this round. RFC 8126 opens the tiers section, because naming the two tiers as Specification Required and Expert Review is the shortest available statement of what they are. The lifecycle section is also the answer to your later comment asking for the transitional phases as a lifecycle with a flow diagram: every state a type can be in, a transition table giving the trigger and the decider for each edge, and a diagram. Two things previously left to inference get stated. Rejection has exactly two grounds, duplication and active harm, and failing the evidence criterion is not one of them. And no transition deletes a record, because a published `predicateType` URI can appear in attestations for years after the type stops being a good idea, so withdrawal leaves a tombstone.

On the criterion section, it becomes "Acceptance and promotion criterion", with the three items outlined in a table giving each item and what a reviewer actually checks before they are defined at length, and E1, E2 and E3 as subsections rather than labelled list items. Your "signed by who? with what?" was the best question in the review, because the document leans on "signed bundle" for several pages and answers it much later and incompletely. Signed by the party that produced the runs, each implementer signing their own, with no requirement that the two parties trust each other, as a DSSE envelope over an in-toto Statement, which is the format this repository already ships tooling for. What the signature does not buy also gets written down, since that is the part most likely to be over-read: it binds runs to a key and to a corpus digest so a run cannot later be attributed to a corpus it was not taken against, and it establishes nothing whatsoever about whether the signer is trustworthy. Key distribution and trust roots stay with verifier policy.

Doors versus gates is a fair hit. Three paths is clearer and less pleased with itself, and the section needs the sentence saying a submission needs only one of the three, which the old wording made the reader work out.

On the risk of someone adopting a fundamentally broken predicate from contrib, this is the most important comment in the review and the section did not cover it. You also caught that the "trust signals are over-read" sentence was too terse to tell what I meant: it was about the opposite risk, someone over-reading vetted, and the risk you identified was absent. The security section needs both halves. Introducing a tier does create a place where a broken type can sit published indefinitely, where today it would either be scrutinised into shape or not listed, and the text should say this ITE creates that exposure. Three things bound it and none eliminates it: rejection for active harm is unchanged, so the dangerous case is still refused; deprecated and withdrawn give the project its first post-listing response, which is a real improvement on a directory where a type once added can only be added; and a machine-readable tier lets a consumer refuse unvetted types by policy rather than by reading a web page, which is the only one of the three that works without a human noticing anything. What stays unmitigated, and should be written down as unmitigated, is a broken contrib type nobody reports, because detecting it is the work the criterion exists to make somebody else do.

Your reading of the index CI job is exactly right and the document should state it outright: the index is generated by a job, the job runs on merge, it does not exist, and this ITE is asking the project to write it. Separating the generator, the workflow and the generated data makes the size of each visible, and the continuing-maintenance assignment ITE-1 requires was missing: generator and workflow maintained by the repository maintainers like any other CI job, per-type artifacts by each type's change controller. I am offering to write and hand over the initial generator and workflow with no expectation of continuing ownership.

On dropping the Testing section I want to do something other than what you asked. I agree a process ITE needs no testing strategy and that what sits under that heading is not one. But ITE-1 lists Testing among the sections an ITE MUST have to be approved as Draft, says an ITE with no testing needs must say so, and warns that submissions not adequately completing a required section will not be approved; ITE-1 itself, also a process ITE, keeps the heading with the single sentence "There are no testing additions related to this ITE." So I would keep the heading, cut it to the sentence ITE-1 asks for, and move the design rationale under it to Reasoning, which is where ITE-1 puts reasoning for design decisions. If you would still rather the section went entirely I will drop it; I did not want to remove a required section on my own initiative when it is you who would be moving this to Draft.

Three defects on the branch as it stands today are mine and are worth naming now, because they are live in what you reviewed. The Reasoning section delegates the alternatives considered, the tooling and the open questions to a `COMPANION.adoc` that I folded into the main document before submitting and that does not exist on the branch; the links are at lines 266 and 385. Two artifact counts disagree with each other, "the five other artifacts named in the companion" at line 327 against "the six artifacts named in the companion" at line 352, and the file that would settle which is right is the missing one. And line 370 says a worked example is "linked below" when it never was.

The fourth is the one I would most like to flag, because it is the same failure this ITE exists to prevent sitting inside the ITE proposing to prevent it. Line 361 says the runnable parts accompany this pull request in a fork and lists seven of them. They are not there. This pull request adds one file, no branch of my fork of ITE carries that tooling, and neither does any branch of my fork of in-toto/attestation. Anyone following that sentence finds nothing, and it is still standing. It should be withdrawn and replaced with what a reader can actually open: a published corpus in the E1 shape at github.com/astrogilda/aee-conformance, and more usefully a verifier written by someone other than me, in a different language, running against it. The first thing the two implementations disagreed about was a bound the specification had never stated: the other implementer and I had independently picked different maximum JSON nesting depths, so for every depth between the two the same bytes were valid evidence to one conformant verifier and malformed to the other. Five further implementations I had written myself agreed with each other at every depth and could never have surfaced it. The bound is now normative in the specification text, which is where E3 asks disagreements to be resolved. It is still not a passing example: the corpus is not shipped with the submission, there is no index record to reach it from, and it is not published as a signed bundle, so it fails E1 as written.

On size, you asked twice for less rather than more, and everything above is net growth. The sections you named both shrink, but the lifecycle section, the security section, the signing subsection, the infrastructure detail, and the alternatives and open questions ITE-1 requires all add. That is an explanation and not a defence, and if you want the Reasoning subsections trimmed or moved back out to a properly committed companion file I am happy to do either; my one worry is recreating the split that produced the dangling links.

On timing, the next community meeting is 4 September. I would like this in a state where the Editors can either approve it as Draft or assign it to whoever they want to own it, and where the tier-name question and the "is Testing required here" question are decided, since both are yours to call. I am not attached to owning this document, which was written because the meeting agreed to file a process ITE and named nobody; if one of the Editors would rather write it, they should take this as material to work from.

Edited 2026-08-20. The earlier version of this comment reported every change above as already pushed, including specific before-and-after word counts and renamed section headings. None of it was on the branch, which still stands at 2edc3955 with the headings, the dangling `COMPANION.adoc` links, the disagreeing artifact counts and the false prototype claim all live. The substance is unchanged and is now written as what the revision will do.

ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
+= ITE-0000: Tiers and an evidence criterion for predicates
+:source-highlighter: pygments
+:toc: preamble
+:toclevels: 2
+
+.Metadata
+[cols="2"]
+|===
+| ITE
+| 0000
```
SantiagoTorres: This is now ITE-12. Can you update the metadata to reflect this?

ITE/0000/README.adoc
```diff
@@ -0,0 +1,405 @@
...
+
+| Title
+| Tiers and an evidence criterion for predicates
+
+| Sponsor
+| link:https://github.com/astrogilda[Sankalp Gilda]
```
SantiagoTorres: The sponsor should be an ITE editor that is sponsoring this. It should either be me or @adityasaky, who have been involved.

@adityasaky any thoughts on this? should I lead it ?

SantiagoTorres:
Oh @astrogilda sorry I thought you had already pushed the changes and I was having a hard time making them. Do push the changes please. I'll take a look at this this weekend :)

astrogilda:
Pushed. The branch is at `9de9111`.

Thirteen of your fifteen comments are in. I left two open because they are yours to call.

Three tier name candidates are live: vetted and contrib as they stand, core and contrib as you suggested, or community from the 7 August meeting. `docs/new_predicate_guidelines.md` already says "our list of vetted predicates", so vetted renames nothing already published, while core renames a concept this ITE was not otherwise touching. Contrib matches the convention you are pointing at, though Debian uses it for packages depending on non-free ones, which carries a faintly second-class sense this ITE spends effort arguing against. Community carries none of that. I have no preference and it is a mechanical substitution at this point, so say which.

You are right that a process ITE needs no testing strategy, and right that what sat under that heading was not one. I kept the heading because ITE-1 requires the section for Draft approval and asks an ITE with no testing needs to say so. It now says that in one sentence and points to the submitter testing requirements. Say the word and I will drop the section.

Both dangling COMPANION.adoc links are gone; the reasoning they delegated to is now in the document. The mismatched artifact counts (five in one section, six in another) are reconciled, and Infrastructure names the artifacts; it no longer points at the companion I had deleted. Finally, Prototype Implementation no longer says the runnable parts accompany this pull request in a fork. They did not and do not. This pull request adds one file, and the section says so in those words. That was the exact failure this ITE exists to prevent, live in the text you reviewed. What replaces it is what a reader can open: the corpus at github.com/astrogilda/aee-conformance, and a verifier written by someone other than me, in a different language, running against it. It is still not a passing example and the text says so, because the corpus does not ship with a predicate submission, no index record reaches it, and it is not published as a signed bundle.

You asked twice for less size, and this round is still net growth. Scope and the abstract shrink. Lifecycle, security, signing and infrastructure all grow, and the alternatives and open questions ITE-1 requires came in from COMPANION.adoc. If you want the Reasoning subsections trimmed, say so; trimming them risks recreating the split that produced the dangling links.

Next community meeting is 4 September. I want this ready for you to approve as Draft or reassign. I am not attached to owning the document.

ITE/0000/README.adoc
astrogilda: It now reads as one paragraph: what sorts a submission into a tier, and the fact that today's bar still lands one in contrib.

ITE/0000/README.adoc
astrogilda: Simplified, and I took that as the instruction for the whole document. Scope is now two lines, one for what this ITE changes and one for what it doesn't.

ITE/0000/README.adoc
astrogilda: Not quite, and that reading is now ruled out. Determinate has its own subsection before first use: two readers implementing from the text alone reach the same accept-or-reject decision.

ITE/0000/README.adoc
astrogilda: RFC 8126 now opens the tiers section, and a new Lifecycle section names the states and transitions. There are two grounds for rejection: duplication and active harm.

ITE/0000/README.adoc
astrogilda: Renamed to "Acceptance and promotion criterion", using your title.

ITE/0000/README.adoc
astrogilda: By whoever produced the runs in the bundle, not by the project and not by a central authority. The signature is a DSSE envelope over an in-toto Statement.

ITE/0000/README.adoc
astrogilda: Done. E1, E2 and E3 are each their own sub-sub-section now, along with the signing rules and the delegated-schema case, so each is reachable from the table of contents.

ITE/0000/README.adoc
astrogilda: Renamed to "Promotion from contrib to vetted". The old heading sounded better than it read.

ITE/0000/README.adoc
astrogilda: Paths throughout; door and gate are gone. The three paths are numbered items under the renamed promotion section.

ITE/0000/README.adoc
astrogilda: Framing it as a lifecycle showed two inconsistencies of mine. There's now a Lifecycle section naming six states, with a flow diagram for the transitions.

ITE/0000/README.adoc
astrogilda: The Security section now carries that risk: a broken type can sit in contrib, listed, indefinitely, and one nobody reports stays unmitigated. The terse phrasing is gone.

ITE/0000/README.adoc
astrogilda: Yes, and it says so now: the index is generated by a CI job that runs on merge, and the job doesn't exist yet. I offer to write it.

ITE/0000/README.adoc
astrogilda: Dropped, both the Testing section and the falsification-conditions block. ITE-1 asks for a testing section only where an ITE includes code, which this one doesn't.

astrogilda:
Thanks Santiago, all sixteen comments are applied with a reply on each, and 58eee06 drops the Testing, Unresolved and Alternatives sections.

Edited 2026-09-12 to name the sections cut.

ITE/0000/README.adoc
astrogilda: It's a mechanical substitution and the Editors should use whichever pair they prefer; I kept vetted because `docs/new_predicate_guidelines.md` already calls the list that. Say the word and I'll swap.

astrogilda:
Pushed two non-normative fixes to ITE/0000/README.adoc; no tier definition is touched. The corpus under Prototype Implementation was renamed on 2026-09-11, so it now carries the current name, agent-evidence-vectors, with its size at a pinned tag. Three References entries had backticks inside the link macro, which AsciiDoc renders as literal characters in the URL, so those three are plain URLs now. Backticks in link text are untouched.

astrogilda:
@SantiagoTorres all sixteen of your comments are applied, with a reply on each.

In that round 58eee06 also dropped the Testing, Unresolved and Alternatives sections.

One call is still yours: the tier names, where vetted and contrib, core and contrib, or community are all live.

The changes-requested review is still the state, so a second look from you is what moves this.

